PR #7, opened 2017-06-05
JVM interface

phi-dbq:

1. Register standard single UDF and composition of UDFs
2. Register UDFs from TensorFlow graph and execute
3. Export JVM Spark instances 
   1. Provide access to SparkSession, SparkContext and SQLContext
   2. Export utility functions for converting various collections

## How is this patch tested?

- [ ] Unit tests
- [ ] Integration tests
- [x] Manual tests


python/sparkdl/utils.py
```diff
@@ -20,38 +21,64 @@
...
 
-ENTRYPOINT_CLASSNAME = "com.databricks.sparkdl.python.PythonInterface"
-
 logger = logging.getLogger('sparkdl')
 
-def _java_api_sql(javaClassName, sqlCtx = None):
-    """
-    Loads the PythonInterface object (lazily, because the spark context needs to be initialized
-    first).
+class JVMAPI(object):
```
MrBago: Sometimes it can be useful to create a namespace, but in this case I'd actually move it into it's own module. It should be cleaner that way. For example within the module you can drop the cls in statements like `cls._curr_sc()._jvm` and use `_curr_sc()._jvm`. You'll also get the benefits of a namespace, if you call the file `jvmapi.py` you'll use `jvmapi.for_class` and so on.

phi-dbq:
closing this as it is duplicate of [#15]

PR #11, opened 2017-06-06
Replace test images with images from Getty's Open Content Program.

MrBago:


sueann:
I'm guessing the tests run? Non-art images are probably better, but I think it's okay as long as the tests run.

PR #26, opened 2017-06-20
Register Keras image model as Spark SQL UDF

phi-dbq:

This PR creates functionalities for users to register Keras image model as Spark SQL UDF.
The user also has the option to provide an image preprocessing function. 

## How is this patch tested?

- [x] Unit tests
- [x] Manual tests


python/sparkdl/utils/jvmapi.py
```diff
@@ -59,6 +59,22 @@ def default():
...
+    Given a sequence of @ordered_udf_names f1, f2, ..., fn
+    Create a pipelined UDF as fn(...f2(f1()))
+    """
+    assert len(ordered_udf_names) > 1, \
+        "must provide more than one ordered udf names"
+    return default().registerPipeline(name, ordered_udf_names)
+
+def registerUDF(name, function_body, schema):
+    """ 
+    Given a sequence of @ordered_udf_names f1, f2, ..., fn
```
thunterdb: is this true?

python/tests/udf/keras_sql_udf_test.py
```diff
@@ -0,0 +1,137 @@
...
+        # Load image with Keras and store it in our image schema
+        JVMAPI.registerUDF('keras_load_spimg', keras_load_spimg, imageSchema)
+        JVMAPI.registerUDF('pil_load_spimg', pil_load_spimg, imageSchema)
+
+        # Register an InceptionV3 model
+        registerKerasImageUDF("iv3_img_pred",
+                              InceptionV3(weights="imagenet"),
+                              keras_load_img)
+
+        SQL = self.session.sql
```
thunterdb: why upper case?

python/tests/udf/keras_sql_udf_test.py
```diff
@@ -0,0 +1,137 @@
...
+        print("df2 = %s" % df2)
+        data2 = df2.collect()
+        assert data2[0].z == 3.0, data2
+
+
+    def test_map_blocks_sql_1(self):
+        data = [Row(x=float(x)) for x in range(5)]
+        df = self.sql.createDataFrame(data)
+        with IsolatedSession() as issn:
+            # The placeholder that corresponds to column 'x'
```
thunterdb: ... as a whole column

python/tests/udf/sql_test.py
```diff
@@ -0,0 +1,127 @@
...
+    return df
+
+
+class SqlUserDefinedFunctionTest(SparkDLTestCase):
+
+    def test_composite_udf(self):
+        df = get_image_paths_df(self.sql)
+        df.show()
+
+        def keras_load_img(fpath):
```
thunterdb: this is all repeated between tests, do not copy.

python/tests/udf/sql_test.py
```diff
@@ -0,0 +1,127 @@
+#
+# Copyright 2017 Databricks, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
```
thunterdb: I am confused, what is the difference with the file above?

python/sparkdl/graph/builder.py
```diff
@@ -75,6 +78,48 @@ def run(self, *args, **kargs):
...
+    def asUDF(self, udf_name, fetches, feed_dict=None, blocked=False, register=True):
+        """ 
+        Make the TensorFlow graph as a SQL UDF 
+
+        :param udf_name: str, name of the SQL UDF 
+        :param fetches: list, output tensors of the graph
+        :param feed_dict: dict, a dictionary that maps graph elements to input values
+        :param blocked: bool, whether the TensorFrame execution should be blocked based or row based
+        :param register: bool, whether this UDF should be registered
+        :return: JVM function handle object
```
thunterdb: this return value is very unhelpful for python users, but I want to propose a way to wrap a scala UDF automatically. No need to change it for now.

Update the doc though to mention that it is of class `... UserDefinedFunction`

python/sparkdl/graph/builder.py
```diff
@@ -75,6 +78,48 @@ def run(self, *args, **kargs):
...
+        """ 
+        # pylint: disable=W0212
+        # TODO: work with registered expansions
+        jvm_builder = JVMAPI.forClass(JVMAPI.MODEL_FACTORY_CLASSNAME)
```
thunterdb: In retrospect, I am not a fan for calling `forClass` because it also sets the default the sql context. Something like `createBuilder` is good, I think.

python/sparkdl/udf/keras_image_model.py
```diff
@@ -0,0 +1,90 @@
...
+        img_arr_reloaded = img_arr_reloaded.astype(np.uint8)
+        return imageArrayToStruct(img_arr_reloaded)
+
+    return udf_impl
+
+def registerKerasImageUDF(udf_name, keras_model_or_file_path, preprocessor=None):
+    """
+    Create an UserDefinedFunction from a Keras model
+
+    registerKerasImageUDF(`udf_name`, "path/to/my/keras/model.h5", `preprocessor`)
```
thunterdb: This is the main function of this file, and it should be well documented. You should use here the pydoc syntax. Ideally, this should be a standalone docstring test, but that requires a bit of setup. You should explain at least how to get a H5 file (or point to the keras doc).

Also, the main usage will probably be a python object, not a file. You can put something simple like what is here: https://keras.io/getting-started/sequential-model-guide/

```
model = Sequential()
model.add(Dense(32, input_dim=784))
```

python/sparkdl/udf/keras_image_model.py
```diff
@@ -0,0 +1,90 @@
...
+
+    :param udf_name: str, name of the UserDefinedFunction
```
thunterdb: the name of the SQL function, once it is registered. If the name exists, it will be overwritten.

python/sparkdl/udf/keras_image_model.py
```diff
@@ -0,0 +1,90 @@
...
+    :param keras_model_file_path: str, path to the HDF5 keras model file
```
thunterdb: do we just accept file paths? I thought we were also accepting models. I recall that there are some serialization issues to consider for that?

python/sparkdl/udf/keras_image_model.py
```diff
@@ -0,0 +1,90 @@
...
+    :param preprocessor: function, optional, 
+    :return: list, image as a DataFrame Row compatible list    .
```
thunterdb: this is not correct?

python/sparkdl/udf/keras_image_model.py
```diff
@@ -0,0 +1,90 @@
...
+
+def registerKerasImageUDF(udf_name, keras_model_or_file_path, preprocessor=None):
+    """
+    Create an UserDefinedFunction from a Keras model
+
+    registerKerasImageUDF(`udf_name`, "path/to/my/keras/model.h5", `preprocessor`)
+
+    :param udf_name: str, name of the UserDefinedFunction
+    :param keras_model_file_path: str, path to the HDF5 keras model file
+    :param preprocessor: function, optional,
```
thunterdb: you should explain what this is, and where it comes from. There should be an example above with inception, so that people understand the necessity. This is just a matter of copy/pasting one of the tests.

.travis.yml
```diff
@@ -16,7 +16,7 @@ cache:
 
 env:
   matrix:
-    - SCALA_VERSION=2.11.8 SPARK_VERSION=2.1.1 SPARK_BUILD="spark-2.1.1-bin-hadoop2.7" SPARK_BUILD_URL="http://d3kbcqa49mib13.cloudfront.net/spark-2.1.1-bin-hadoop2.7.tgz"
+    - SCALA_VERSION=2.11.8 SPARK_VERSION=2.1.1 SPARK_BUILD="spark-${SPARK_VERSION}-bin-hadoop2.7" SPARK_BUILD_URL="http://d3kbcqa49mib13.cloudfront.net/spark-${SPARK_VERSION}-bin-hadoop2.7.tgz"
```
thunterdb: I did not know you could chain variables like this.

python/tests/udf/sql_test.py
phi-dbq: Oh, my bad ~ it is a merge mistake

python/sparkdl/graph/builder.py
```diff
@@ -75,6 +78,48 @@ def run(self, *args, **kargs):
         """
         return self.sess.run(*args, **kargs)
 
+    def asUDF(self, udf_name, fetches, feed_dict=None, blocked=False, register=True):
```
sueann: Is this a comment suggesting action for this PR, or a comment to leave a comment here? 😅

python/sparkdl/graph/builder.py
```diff
@@ -75,6 +78,48 @@ def run(self, *args, **kargs):
...
+        """ 
+        Make the TensorFlow graph as a SQL UDF 
+
+        :param udf_name: str, name of the SQL UDF 
+        :param fetches: list, output tensors of the graph
+        :param feed_dict: dict, a dictionary that maps graph elements to input values
+        :param blocked: bool, whether the TensorFrame execution should be blocked based or row based
```
sueann: not sure if there is a description that will be clear enough lol. do we actually support blocking for these UDFs?

sueann: actually, what does it mean to have blocking for a SQL UDF?

python/sparkdl/udf/keras_image_model.py
```diff
@@ -0,0 +1,90 @@
...
+from ..graph.builder import GraphFunction, IsolatedSession
+from ..graph.pieces import buildSpImageConverter, buildFlattener
+from ..image.imageIO import imageSchema
+from ..utils import jvmapi as JVMAPI
+
+logger = logging.getLogger('sparkdl')
+
+def _serialize_and_reload_with(preprocessor):
+    """
+    Load a preprocessor function (image_file_path => image_tensor)
```
sueann: More like "Returns a function that loads and preprocesses images using the given preprocessor  (image_file_path => image_tensor).

python/sparkdl/udf/keras_image_model.py
sueann: Also specify what the input and output of the preprocessor should be (file path string => numpy array)

python/sparkdl/udf/keras_image_model.py
```diff
@@ -0,0 +1,90 @@
...
+            "expect preprocessor to return a numpy array"        
+        img_arr_reloaded = img_arr_reloaded.astype(np.uint8)
+        return imageArrayToStruct(img_arr_reloaded)
+
+    return udf_impl
+
+def registerKerasImageUDF(udf_name, keras_model_or_file_path, preprocessor=None):
+    """
+    Create an UserDefinedFunction from a Keras model
+
```
sueann: Also need to say what the input and output column types for the resulting UDF are.

python/tests/udf/sql_test.py
```diff
@@ -0,0 +1,127 @@
...
+                              keras_load_img)
+
+        SQL = self.session.sql
+
+        # Choice 1: manually chain the functions in SQL
+        df1 = SQL("select iv3_img_pred(keras_load_spimg(fpath)) as preds from _test_image_paths_df")
+        preds1 = np.array(df1.select("preds").rdd.collect())
+
+        # Choice 2: build a pipelined UDF and directly use it in SQL
+        JVMAPI.registerPipeline("load_img_then_iv3_pred", ["keras_load_spimg", "iv3_img_pred"])
```
sueann: one potential optimization:
For this use case where the input dataframe to the SQL UDF contains file paths, instead of having to register a pipeline of load => predict, we could have a version of registerKerasUDF that natively handles file path input columns. This way we could avoid doing:
file path => load file => write file => load file using the given preprocessor => predict.

sueann: if that sounds interesting, let's talk in person a little before doing it

python/sparkdl/transformers/tf_image.py
```diff
@@ -35,17 +35,18 @@ class TFImageTransformer(Transformer, HasInputCol, HasOutputCol):
     Applies the Tensorflow graph to the image column in DataFrame.
 
     Restrictions of the current API:
+
```
phi-dbq: This blank line is needed or sphinx won't parse the lists correctly

python/sparkdl/graph/builder.py
```diff
@@ -67,17 +69,101 @@ def __exit__(self, *args):
...
         """
         return self.sess.run(*args, **kargs)
 
+    def asUDF(self, udf_name, fetches, feeds_to_fields_map=None, blocked=False, register=True):
+        """
+        Register the graph in this session as a Spark SQL UserDefinedFunction.
+
+        The following example creates a UDF that takes the input
+        from a DataFrame column named 'image_col' and produce some random prediction.
+
```
thunterdb: here is how to make the code testable:
https://thomas-cokelaer.info/tutorials/sphinx/doctest.html

python/sparkdl/graph/builder.py
```diff
@@ -67,17 +69,101 @@ def __exit__(self, *args):
...
+        .. code-block:: python
+
+            import sparkdl.graph.utils as tfx
+
+            with IsolatedSession(using_keras=True) as issn:
```
thunterdb: do you need to isolate?

python/sparkdl/graph/builder.py
```diff
@@ -67,17 +69,101 @@ def __exit__(self, *args):
...
+                model = Sequential()
+                model.add(Flatten(input_shape=(640,480,3)))
+                model.add(Dense(units=64))
```
thunterdb: just put one dense layer for the purpose of example

python/sparkdl/graph/builder.py
thunterdb: hmm ok, we use it after that

python/sparkdl/graph/builder.py
```diff
@@ -67,17 +69,101 @@ def __exit__(self, *args):
...
+                model.add(Activation('relu'))
+                model.add(Dense(units=10))
+                model.add(Activation('softmax'))
+                # Initialize the variables
+                init_op = tf.global_variables_initializer()
```
thunterdb: is it required?

python/sparkdl/graph/builder.py
```diff
@@ -67,17 +69,101 @@ def __exit__(self, *args):
...
+                issn.run(init_op)
+                # Export the model as UDF
+                issn.asUDF('my_keras_model_udf',
+                           model.outputs,
+                           {tfx.op_name(issn.graph, model.inputs[0]): 'image_col'})
```
thunterdb: can you directly get the name from the layer? It should be something simple?

python/sparkdl/udf/keras_image_model.py
```diff
@@ -0,0 +1,146 @@
...
+    .. code-block:: python
+
+        registerKerasImageUDF("udf_name", "path/to/my/keras/model.h5", preprocessor)
+
+    Or, we can provide the model
+
+    .. code-block:: python
+
+        from keras.applications import InceptionV3
```
thunterdb: import also the preprocessor

thunterdb: this is the most common use case.

python/sparkdl/graph/builder.py
thunterdb: I had forgotten that this is just for tensorflow. The example here should be minimal, like the test below:

```python
		 +        with IsolatedSession() as issn:
 +            # The placeholder that corresponds to column 'x' as a whole column
 +            x = tf.placeholder(tf.double, shape=[], name="x")
 +            # The output that adds 3 to x
 +            z = tf.add(x, 3, name='z')
 +            # Let's register these computations in SQL.
 +            issn.asUDF("map_rows_sql_1", [z])
```

python/sparkdl/graph/builder.py
thunterdb: actually see comment above.

python/sparkdl/graph/builder.py
thunterdb: see comment above about this example

python/sparkdl/graph/builder.py
thunterdb: see comment above about this example

python/sparkdl/graph/builder.py
```diff
@@ -67,17 +69,101 @@ def __exit__(self, *args):
...
+        to the underlying tf.Session object to perform
         one step of graph computation.
 
         All the parameters are defined according to `tf.Session.run`
-        Reference: https://www.tensorflow.org/api_docs/python/tf/Session#run        
+        Reference: https://www.tensorflow.org/api_docs/python/tf/Session#run
         """
         return self.sess.run(*args, **kargs)
 
+    def asUDF(self, udf_name, fetches, feeds_to_fields_map=None, blocked=False, register=True):
```
thunterdb: I had missed that before: this function does not use a full session, just the current graph of computation. Instead of making it a method to the isolated session, make it a top-level function with the first argument being a graph. This function is meant to be low level, and I want to move it to TensorFrames eventually.

python/sparkdl/graph/builder.py
phi-dbq: Our code examples require setting up spark and tensorflow/keras for testing. Maybe we can provide them as examples and import them in the doc?

python/sparkdl/graph/builder.py
phi-dbq: Looks like we have to explicitly call `model.build()` to have the `model.input_names` available. https://github.com/fchollet/keras/blob/master/keras/models.py#L535

python/sparkdl/graph/builder.py
sueann: Let's just put it in the README.md and the databricks guide.

python/sparkdl/graph/spark_utils.py
```diff
@@ -0,0 +1,108 @@
+#
```
sueann: Why is this under graph/ and the keras one under udf/ ? Is this one not meant to be exposed?

python/sparkdl/graph/spark_utils.py
```diff
@@ -0,0 +1,108 @@
...
+                                For now, only the top-level fields in a dataframe are supported.
+
+                                .. note:: For any placeholder that is
+                                          not specified in the feed dictionary,
+                                          the name of the input column is assumed to be
+                                          the same as that of the placeholder.
+
+    :param blocked: bool, if set to True, the TensorFrames will execute the function
+                    over blocks/batches of rows. This should provide better performance.
+                    Otherwise, the function is applied to individual rows
```
sueann: why would you ever have it not blocked? should we just make it always blocked and remove the argument?

python/sparkdl/udf/keras_image_model.py
```diff
@@ -0,0 +1,146 @@
...
+from sparkdl.graph.spark_utils import makeGraphUDF
+from sparkdl.image.imageIO import imageSchema
+from sparkdl.utils import jvmapi as JVMAPI
+
+logger = logging.getLogger('sparkdl')
+
+def registerKerasImageUDF(udf_name, keras_model_or_file_path, preprocessor=None):
+    """
+    Create a Keras image model as a Spark SQL UDF.
+    The function takes a column (formatted in :py:const:`sparkdl.image.imageIO.imageSchema`)
```
sueann: function -> UDF

python/sparkdl/udf/keras_image_model.py
```diff
@@ -0,0 +1,146 @@
...
+    and produce a prediction as a probability over of a set of known categories.
```
sueann: nit: produce -> produces

the output is any arbitrary vector right? doesn't have to be a probability vector. I forget if the return type is a list or a MLlib Vector, but should specify here.

thunterdb: Good point, I have also forgotten.

python/sparkdl/graph/spark_utils.py
thunterdb: do we want to expose it here? I would prefer to keep it private-ish for the time being, because this should belong to TensorFrames: it calls all sorts of internals and it is not specific to deep learning.

sueann: we should add that in a comment here then. Maybe rename the module to `tensorframes_udf` if we really want this to be in tensorframes eventually. `spark_utils` sounds like it's got a bunch of helper functions around spark.

phi-dbq: I intend to consider things under graph/ as more like developer tools whereas things under udf/ are baked UDFs that end users can use.

python/sparkdl/graph/spark_utils.py
sueann: from @thunterdb : two scenarios under which you would use blocked = False:
1/ your tensorflow graph tensors do not specify the "batch" dimension (the first dimension in the shape)
2/ rows have varying sizes, e.g. sparse vectors

let's document this here.

phi-dbq: If `blocked = true`, the graph provided by the user should [accept batch input](https://github.com/databricks/spark-deep-learning/blob/master/src/main/scala/org/tensorframes/impl/SqlOps.scala#L116), e.g. `input_shape=(None, W, H, C)`. Our current `Spimage` allows images to have different width and heights, thus batching might not be able to directly apply.

python/sparkdl/udf/keras_image_model.py
phi-dbq: Yeah, I think the output is just an arbitrary vector (`WrappedArray`).

python/sparkdl/graph/spark_utils.py
sueann: Ah yeah, for the Keras udf registration functinon we do need to talk about `SpImage`. It might fail if the images have different widths and heights, but it could succeed if they all had the same widths and heights right?

python/sparkdl/udf/keras_image_model.py
```diff
@@ -0,0 +1,147 @@
...
+from sparkdl.utils import jvmapi as JVMAPI
+
+logger = logging.getLogger('sparkdl')
+
+def registerKerasImageUDF(udf_name, keras_model_or_file_path, preprocessor=None):
+    """
+    Create a Keras image model as a Spark SQL UDF.
+    The UDF takes a column (formatted in :py:const:`sparkdl.image.imageIO.imageSchema`)
+    and produces the output of the given Keras model (which is a real valued score vector 
+    over predefined categories, e.g. `Inception V3 <https://keras.io/applications/#inceptionv3]>`_)
```
sueann: this also can output an arbitrary vector no?

sueann: by that i mean the "categories" part is not necessary. for example, the input model may not actually output a classification probability vector but a transformed image, no? In such a case, the UDF returns a flat vector, not a `SpImage`, right? (For KerasImageFileTransformer, there is an option to return a SpImage or a 1-d vector).

phi-dbq: Yes, you are right. The model dictates the output.

python/sparkdl/graph/spark_utils.py
phi-dbq: Yes. To apply a vanilla Keras Inception model, the width and height have to be fixed. For any popular image mode, the input is expected to be of a fixed size, unless reshaping to canonical size from provided width and height is part of the model graph, like we have in the transformers.

python/sparkdl/graph/spark_utils.py
thunterdb: +1 for renaming to `tensorframes_udf`

python/sparkdl/graph/spark_utils.py
thunterdb: I will need to document this part, but TensorFrames expects a first analysis pass that adds some metadata with the shape info in order to work correctly with blocks. It does not matter in the case of simple inputs like scalars. I want to make the input of tensorframes more robust to mis-shaped batches, but I am waiting to see the sort of use cases that people come up with before making changes here.

phi-dbq:
@sueann @thunterdb any additional comments?

phi-dbq:
@thunterdb @MrBago I think adding the following to nose would remove most coverage stats from external packages
`--with-coverage --cover-package sparkdl`

sueann:
lgtm. let's merge this morning.

PR #30, opened 2017-07-03
Adding generic support for Keras pre-trained models

sueann:
DeepImageFeaturizer/Predictor currently has InceptionV3. We want to add other pre-trained models available from Keras. Here is a simple clean way to add them. This PR adds the light wrappers around models and tests for the existing InceptionV3 model. In the next PR we can add all the models and some control over how extensively to run the tests.

The code to use the various Keras pre-trained model is basically the same so it is captured here under KerasApplicationModel.getModelData. See [https://keras.io/applications](https://keras.io/applications).

The `ModelData` interface between the classes is a bit fragile, but seems okay for now...

### DIFF
--- a/python/sparkdl/transformers/keras_applications.py
+++ b/python/sparkdl/transformers/keras_applications.py
@@ -20,8 +20,7 @@ from keras.applications import inception_v3, resnet50, vgg16, vgg19, xception
 from keras.applications.imagenet_utils import decode_predictions
 import tensorflow as tf
 
-from sparkdl.transformers.utils import (
-    imageInputPlaceholder, InceptionV3Constants)
+from sparkdl.transformers.utils import (imageInputPlaceholder, InceptionV3Constants)
 
 
 KERAS_APPLICATION_MODELS = set(["InceptionV3"]) #, "ResNet50", "Xception", "VGG16", "VGG19"])
@@ -86,6 +85,13 @@ class KerasApplicationModel:
         """
         pass
 
+    @abstractmethod
+    def testKerasModel(self):
+        """
+        For testing, the keras model object to compare to.
+        """
+        pass
+
 
 class InceptionV3Model(KerasApplicationModel):
     def preprocess(self, inputImage):
@@ -101,4 +107,5 @@ class InceptionV3Model(KerasApplicationModel):
     def numOutputFeatures(self):
         return InceptionV3Constants.NUM_OUTPUT_FEATURES
 
-
+    def testKerasModel(self):
+        return inception_v3.InceptionV3()
--- a/python/sparkdl/transformers/keras_applications.py
+++ b/python/sparkdl/transformers/keras_applications.py
@@ -16,18 +16,18 @@
 from abc import ABCMeta, abstractmethod
 
 import keras.backend as K
-from keras.applications import inception_v3, resnet50, vgg16, vgg19, xception
+from keras.applications import inception_v3
 from keras.applications.imagenet_utils import decode_predictions
 import tensorflow as tf
 
 from sparkdl.transformers.utils import (imageInputPlaceholder, InceptionV3Constants)
 
 
-KERAS_APPLICATION_MODELS = set(["InceptionV3"]) #, "ResNet50", "Xception", "VGG16", "VGG19"])
+KERAS_APPLICATION_MODELS = set(["InceptionV3"])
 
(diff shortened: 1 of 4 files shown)


python/sparkdl/transformers/named_image.py
```diff
@@ -14,7 +14,7 @@
 #
 
 import keras.backend as K
-from keras.applications import InceptionV3, inception_v3
+from keras.applications import InceptionV3, inception_v3, ResNet50, resnet50
```
sueann: we don't need ResNet50 here yet -- will clean up

sueann:
Sync'ed with @MrBago on how to make the testing lightweight once we add in 4 more models. There are a couple of ways to handle it, and I will do it in the next PR when we actually add in (one or more of) those models.

python/sparkdl/transformers/keras_applications.py
```diff
@@ -0,0 +1,111 @@
...
+        """
+        pass
+
+
+class InceptionV3Model(KerasApplicationModel):
+    def preprocess(self, inputImage):
+        return inception_v3.preprocess_input(inputImage)
+
+    def model(self, preprocessed, featurize):
+        return inception_v3.InceptionV3(input_tensor=preprocessed, weights="imagenet",
```
phi-dbq: I wonder since we are passing the model, can we also use `model.input.shape` to get the shape?

python/sparkdl/transformers/keras_applications.py
```diff
@@ -0,0 +1,111 @@
...
+        raise ValueError("%s is not implemented but is in the supported models list: %s" %
+                         name, str(KERAS_APPLICATION_MODELS))
+
+
+class KerasApplicationModel:
+    __metaclass__ = ABCMeta
+
+    def getModelData(self, featurize):
+        sess = tf.Session()
+        with sess.as_default():
```
phi-dbq: We should probably add a documentation that Keras uses the default session/graph, so the `sess.as_default` will attach the model to the session we are using.

python/sparkdl/transformers/keras_applications.py
```diff
@@ -0,0 +1,111 @@
...
+import tensorflow as tf
+
+from sparkdl.transformers.utils import (imageInputPlaceholder, InceptionV3Constants)
+
+
+KERAS_APPLICATION_MODELS = set(["InceptionV3"]) #, "ResNet50", "Xception", "VGG16", "VGG19"])
+
+
+"""
+Essentially a factory function for getting the correct KerasApplicationModelSession class
```
phi-dbq: `KerasApplicationModelSession `?

python/sparkdl/transformers/keras_applications.py
```diff
@@ -0,0 +1,111 @@
...
+                                        include_top=(not featurize))
+
+    def inputShape(self):
+        return InceptionV3Constants.INPUT_SHAPE
+
+    def numOutputFeatures(self):
+        return InceptionV3Constants.NUM_OUTPUT_FEATURES
+
+    def testKerasModel(self):
+        return inception_v3.InceptionV3()
```
phi-dbq: Will this create a new `InceptionV3` model every time we call this function? 
Shall we also pass the weights argument (or not since keras by default uses "imagenet")

python/sparkdl/transformers/keras_applications.py
sueann: I don't think we need to comment on it since that is the default behavior of keras.

python/sparkdl/transformers/keras_applications.py
sueann: Ah I renamed the class. Will fix the comment.

python/sparkdl/transformers/keras_applications.py
sueann: It will, but in the tests we don't instantiate it more than once. We can cache stuff in the class if it becomes useful. I agree it's clearer to add "imagenet" for weights though.

python/sparkdl/transformers/keras_applications.py
sueann: doesn't seem to work... `ValueError: New image size should have for [hight, width] but got (?, ?, ?, 3)`

thunterdb:
It looks great, thanks!

PR #32, opened 2017-07-07
[ML-2733] Adding Xception to DeepLearningPredictor/Featurizer

sueann:
Adding Xception model to DeepLearningPredictor/Featurizer. Was going to add ResNet50 but ran into some issues - will fix that subsequently. Also didn't add control flow for testing only subsets of models yet since there are only two.

### DIFF
--- a/python/sparkdl/transformers/keras_applications.py
+++ b/python/sparkdl/transformers/keras_applications.py
@@ -38,10 +38,6 @@ def getKerasApplicationModel(name):
         return InceptionV3Model()
     elif name == "ResNet50":
         return ResNet50Model()
-    # elif name == "VGG16":
-    #     return VGG16Model()
-    # elif name == "VGG19":
-    #     return VGG19Model()
     elif name == "Xception":
         return XceptionModel()
     else:
--- a/python/sparkdl/transformers/keras_applications.py
+++ b/python/sparkdl/transformers/keras_applications.py
@@ -16,15 +16,14 @@
 from abc import ABCMeta, abstractmethod
 
 import keras.backend as K
-from keras.applications import inception_v3
+from keras.applications import inception_v3, resnet50, vgg16, vgg19, xception
 from keras.applications.imagenet_utils import decode_predictions
 import tensorflow as tf
 
 from sparkdl.transformers.utils import (imageInputPlaceholder, InceptionV3Constants)
 
 
-KERAS_APPLICATION_MODELS = set(["InceptionV3"])
-
+KERAS_APPLICATION_MODELS = set(["InceptionV3", "ResNet50", "VGG16", "VGG19", "Xception"])
 
 """
 Essentially a factory function for getting the correct KerasApplicationModel class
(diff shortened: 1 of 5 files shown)


python/sparkdl/transformers/keras_applications.py
```diff
@@ -16,30 +16,33 @@
...
-                         str(KERAS_APPLICATION_MODELS))
+        raise ValueError("%s is not a supported model. Supported models: %s" %
+                         (name, str(KERAS_APPLICATION_MODELS)))
 
     if name == "InceptionV3":
         return InceptionV3Model()
+    elif name == "ResNet50":
+        return ResNet50Model()
+    elif name == "Xception":
+        return XceptionModel()
```
MrBago: We can use a dictionary for `KERAS_APPLICATION_MODELS`, something like:
```
KERAS_APPLICATION_MODELS = {"ResNet50": ResNet50Model,
...
}
```

that way this block becomes:

```
try:
    return KERAS_APPLICATION_MODELS[name]()
except KeyError:
   raise ValueError(...)
```
.

python/sparkdl/transformers/keras_applications.py
```diff
@@ -70,15 +73,14 @@ def model(self, preprocessed, featurize):
...
         """
-        For testing
+        For testing, the keras preprocess function for the model.
         """
-        pass
+        return self.preprocess(inputImage)
 
     @abstractmethod
-    def testKerasModel(self):
+    def testKerasModel(self, include_top):
```
MrBago: How do you feel about making these test methods "private"? I think we should make it clear they are not part of the public API.

MrBago:
I just looked over the test errors in travis, is this a resource issue (not enough time/memory on travis)?

python/sparkdl/transformers/named_image.py
```diff
@@ -29,12 +29,13 @@
 from sparkdl.transformers.tf_image import TFImageTransformer
 
 
-SUPPORTED_MODELS = ["InceptionV3"]
+SUPPORTED_MODELS = ["InceptionV3", "ResNet50", "VGG16", "VGG19", "Xception"]
```
sueann: will clean this up

python/sparkdl/transformers/keras_applications.py
sueann: Hm... but then they are not private anyway because of how they are called in tests... ?? Is it weird to make it look private and use it outside the class?

MrBago:
I think in tests it's ok, Scala lets you use `invokePrivate` in tests and
in python I've always encouraged people to write tests for their private
methods so I think accessing X._private from XTestSuite is reasonable.

sueann:
I split out the tests into their own files because locally I was seeing errors with SparkContext again due to multiple test classes being in one file. I am not sure if that's why travis tests are failing, but we'll see.

python/sparkdl/transformers/keras_applications.py
```diff
@@ -16,30 +16,23 @@
...
-    if name == "InceptionV3":
-        return InceptionV3Model()
-    else:
-        raise ValueError("%s is not implemented but is in the supported models list: %s" %
-                         name, str(KERAS_APPLICATION_MODELS))
+    try:
+        return KERAS_APPLICATION_MODELS[name]()
+    except KeyError:
+        raise ValueError("%s is not a supported model. Supported models: %s" %
+                         (name, str(KERAS_APPLICATION_MODELS)))
```
MrBago: Should this be `str(KERAS_APPLICATION_MODELS.keys())`?

MrBago: or maybe even, `", ".join(KERAS_APPLICATION_MODELS.keys())`.

sueann: ah yes. clearly not in a test path......

MrBago:
We should add assertRaises tests, we haven't been good about that.

python/sparkdl/transformers/keras_applications.py
```diff
@@ -16,30 +16,23 @@
 from abc import ABCMeta, abstractmethod
 
 import keras.backend as K
-from keras.applications import inception_v3
+from keras.applications import inception_v3, xception
 from keras.applications.imagenet_utils import decode_predictions
```
phi-dbq: unused import

python/sparkdl/transformers/named_image.py
```diff
@@ -222,8 +224,8 @@ def _transform(self, dataset):
 
 def _buildTFGraphForName(name, featurize):
     if name not in keras_apps.KERAS_APPLICATION_MODELS:
-        raise ValueError("%s is not a supported model. Supported models: %s" % name,
-                         str(KERAS_APPLICATION_MODELS))
+        raise ValueError("%s is not a supported model. Supported models: %s" %
+                         (name, str(KERAS_APPLICATION_MODELS)))
```
phi-dbq: `keras_app.KERAS_APPLICATION_MODELS`?

PR #34, opened 2017-07-10
Report only coverage information of our package

phi-dbq:
This should suppress test coverage log for external packages.

Documentation
http://nose.readthedocs.io/en/latest/plugins/cover.html#cmdoption-cover-package

Now the Travis CI log is much shorter (https://travis-ci.org/databricks/spark-deep-learning/jobs/252104462)

### DIFF
--- a/python/tests/tests.py
+++ b/python/tests/tests.py
@@ -14,6 +14,7 @@
 #
 
 import sys
+import sparkdl
 
 if sys.version_info[:2] <= (2, 6):
     try:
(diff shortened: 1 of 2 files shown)


python/tests/tests.py
```diff
@@ -14,6 +14,7 @@
 #
 
 import sys
+import sparkdl
```
phi-dbq: The coverage report complains if the `sparkdl` package is not imported.

PR #35, opened 2017-07-11
Porting Keras Estimator API and Reference Implementation

phi-dbq:

Creating a Spark MLlib Estimator API for Keras models, with a reference implementation.
It provides a taste of how to ingest Image from URI in a `DataFrame` and use them to train a Keras model.

The changes consist of these components.
1. Extracted a few Params types for Keras Transformers/Estimators.
2. Keras utilities 
    - Serialization: model <=> hdf5 <=> bytes (for broadcast)
    - Check avaialble Keras options (optimizers, loss functions, etc.)
3. Keras Estimator.

## How is this patch tested?

- [x] Unit tests
- [x] Manual tests

### DIFF
--- a/python/sparkdl/estimators/keras_image_file_estimator.py
+++ b/python/sparkdl/estimators/keras_image_file_estimator.py
@@ -135,7 +135,7 @@ class KerasImageFileEstimator(Estimator, HasInputCol, HasInputImageNodeName,
         Fits a model to the input dataset with optional parameters.
 
         .. warning:: This returns the byte serialized HDF5 file for each model to the driver.
-                     If the model file is way too large, the driver might go out-of-memory.
+                     If the model file is large, the driver might go out-of-memory.
                      As we cannot assume the existence of a sufficiently large (and writable)
                      file system, users are advised to not train too many models in a single
                      Spark job.
--- a/python/sparkdl/estimators/keras_image_file_estimator.py
+++ b/python/sparkdl/estimators/keras_image_file_estimator.py
@@ -21,6 +21,7 @@ import logging
 import numpy as np
 
 from pyspark.ml import Estimator
+import pyspark.ml.linalg as spla
 from pyspark.ml.param import Param, Params, TypeConverters
 
 from sparkdl.image.imageIO import imageStructToArray
@@ -52,30 +53,22 @@ class KerasImageFileEstimator(Estimator, HasInputCol, HasOutputCol, HasLabelCol,
                  typeConverter=TypeConverters.toString)
 
     @keyword_only
-    def __init__(self, inputCol=None, outputCol=None, modelFile=None,
-                 labelCol=None, labelCardinality=None, isOneHotLabel=None,
-                 imageLoader=None, outputMode="vector", optimizer=None, loss=None,
-                 kerasFitParams=None):
+    def __init__(self, inputCol=None, outputCol=None, outputMode="vector", labelCol=None,
+                 modelFile=None, imageLoader=None, optimizer=None, loss=None, kerasFitParams=None):
         """
-        __init__(self, inputCol=None, outputCol=None, modelFile=None,
-                 labelCol=None, labelCardinality=None, isOneHotLabel=None,
-                 imageLoader=None, outputMode="vector", optimizer=None, loss=None,
-                 kerasFitParams=None):
+        __init__(self, inputCol=None, outputCol=None, outputMode="vector", labelCol=None,
+                 modelFile=None, imageLoader=None, optimizer=None, loss=None, kerasFitParams=None)
         """
         super(KerasImageFileEstimator, self).__init__()
         kwargs = self._input_kwargs
         self.setParams(**kwargs)
 
     @keyword_only
-    def setParams(self, inputCol=None, outputCol=None, modelFile=None,
-                  labelCol=None, labelCardinality=None, isOneHotLabel=None,
-                  imageLoader=None, outputMode="vector", optimizer=None, loss=None,
-                  kerasFitParams=None):
+    def setParams(self, inputCol=None, outputCol=None, outputMode="vector", labelCol=None,
+                  modelFile=None, imageLoader=None, optimizer=None, loss=None, kerasFitParams=None):
         """
-        setParams(self, inputCol=None, outputCol=None, modelFile=None,
-                  labelCol=None, labelCardinality=None, isOneHotLabel=None,
-                  imageLoader=None, outputMode="vector", optimizer=None, loss=None,
-                  kerasFitParams=None):
+        setParams(self, inputCol=None, outputCol=None, outputMode="vector", labelCol=None,
+                  modelFile=None, imageLoader=None, optimizer=None, loss=None, kerasFitParams=None)
         """
         kwargs = self._input_kwargs
         self._set(**kwargs)
@@ -126,9 +119,9 @@ class KerasImageFileEstimator(Estimator, HasInputCol, HasOutputCol, HasLabelCol,
         model = kmutil.bytes_to_model(modelBytesBc.value)
 
         model.compile(optimizer=estimator.getOptimizer(), loss=estimator.getLoss())
-        fit_params = estimator.getKerasFitParams()
+        _fit_params = estimator.getKerasFitParams()
 
-        model.fit(x=features, y=labels, **fit_params)
+        model.fit(x=features, y=labels, **_fit_params)
         return kmutil.model_to_bytes(model)
 
     def _getNumpyFeaturesAndLabels(self, dataset):
@@ -142,44 +135,34 @@ class KerasImageFileEstimator(Estimator, HasInputCol, HasOutputCol, HasLabelCol,
         localFeatures = []
         localLabels = []
         is_with_labels = label_col is not None
+
+        if is_with_labels:
+            label_schema = image_df.schema[label_col]
+            assert isinstance(label_schema.dataType, spla.VectorUDT), \
+                "must encode labels in one-hot vector format"
+
         for row in image_df.collect():
             spimg = row[tmp_image_col]
             features = imageStructToArray(spimg)
             localFeatures.append(features)
             if is_with_labels:
-                _label = int(row[label_col])
-                localLabels.append(_label)
+                try:
+                    _keras_label = row[label_col].array
+                except ValueError:
+                    raise ValueError("Cannot extract encoded label array")
+                localLabels.append(_keras_label)
 
         if len(localFeatures) == 0:
             raise ValueError("Given empty dataset!")
 
         # We must reshape input data to form to the same size so as to be stacked
         X = np.stack(localFeatures, axis=0)
-        if not is_with_labels:
-            return X, None
-
-        y_category = np.asarray(localLabels).ravel()
-        if self.isDefined(self.labelCardinality):
-            category_field_size = self.getLabelCardinality()
-        else:
-            category_field_size = np.max(y_category) + 1
-            warn_msg = 'inferring label cardinality from training labels {}'
-            warn_msg = warn_msg.format(category_field_size)
-            logger.warning(warn_msg)
-
-        # We also need to convert them to one-hot encoding
-        if self.getIsOneHotLabel():
-            y = y_category
-            assert y.shape[1] == category_field_size, \
-                "label categories must conform {} != {}".format(y.shape[1], category_field_size)
-        else:
-            local_train_set_size = y_category.shape[0]
-            y_one_hot = np.zeros((local_train_set_size, category_field_size))
-            y_one_hot[np.arange(local_train_set_size), y_category] = 1
-            y = y_one_hot
-            info_msg = 'shape of the training labels {}'.format(y.shape)
-            logger.warning(info_msg)
-
+        y = None
+        if is_with_labels:
+            if not localLabels:
+                raise ValueError("Failed to load any labels from dataset, but labels are required")
+            y = np.stack(localLabels, axis=0)
+            assert y.shape[0] == X.shape[0], "must have same amount of features and labels"
         return X, y
 
     def _collectModels(self, kerasModelsBytesRDD):
--- a/python/sparkdl/estimators/keras_image_file_estimator.py
+++ b/python/sparkdl/estimators/keras_image_file_estimator.py
@@ -1,0 +1,272 @@
+#
+# Copyright 2017 Databricks, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# pylint: disable=protected-access
+from __future__ import absolute_import, division, print_function
+
+import logging
+import numpy as np
+
+from pyspark.ml import Estimator
+from pyspark.ml.param import Param, Params, TypeConverters
+
+from sparkdl.image.imageIO import imageStructToArray
+from sparkdl.param import (
+    keyword_only, HasInputCol, HasOutputCol, HasLabelCol,
+    CanLoadImage, HasKerasModel, HasTensorOutput)
+from sparkdl.transformers.keras_image import KerasImageFileTransformer
+import sparkdl.utils.jvmapi as JVMAPI
+import sparkdl.utils.keras_model as kmutil
+
+__all__ = ['KerasImageFileEstimator']
+
+logger = logging.getLogger('sparkdl')
+
+class KerasImageFileEstimator(Estimator, HasInputCol, HasOutputCol, HasLabelCol,
+                              CanLoadImage, HasKerasModel, HasTensorOutput):
+    """
+    Build a Estimator from a Keras model.
+    """
+
+    # TODO(phi-dbq): allow users to pass a Keras optimizer object
+    #                so that they can configure learning rate, momentum, etc.
+    optimizer = Param(Params._dummy(), "optimizer",
+                      "named Keras optimizer (str), please refer to https://keras.io/optimizers",
+                      typeConverter=TypeConverters.toString)
+
+    loss = Param(Params._dummy(), "loss",
+                 "named Keras loss (str), please refer to https://keras.io/losses",
+                 typeConverter=TypeConverters.toString)
+
+    @keyword_only
+    def __init__(self, inputCol=None, outputCol=None, modelFile=None,
+                 labelCol=None, labelCardinality=None, isOneHotLabel=None,
+                 imageLoader=None, outputMode="vector", optimizer=None, loss=None,
+                 kerasFitParams=None):
+        """
+        __init__(self, inputCol=None, outputCol=None, modelFile=None,
+                 labelCol=None, labelCardinality=None, isOneHotLabel=None,
+                 imageLoader=None, outputMode="vector", optimizer=None, loss=None,
+                 kerasFitParams=None):
+        """
+        super(KerasImageFileEstimator, self).__init__()
+        kwargs = self._input_kwargs
+        self.setParams(**kwargs)
+
+    @keyword_only
+    def setParams(self, inputCol=None, outputCol=None, modelFile=None,
+                  labelCol=None, labelCardinality=None, isOneHotLabel=None,
+                  imageLoader=None, outputMode="vector", optimizer=None, loss=None,
+                  kerasFitParams=None):
+        """
+        setParams(self, inputCol=None, outputCol=None, modelFile=None,
+                  labelCol=None, labelCardinality=None, isOneHotLabel=None,
+                  imageLoader=None, outputMode="vector", optimizer=None, loss=None,
+                  kerasFitParams=None):
+        """
+        kwargs = self._input_kwargs
+        self._set(**kwargs)
+        return self
+
+    def setLabelCol(self, value):
+        return self._set(labelCol=value)
+
+    def getLabelCol(self):
+        return self.getOrDefault(self.labelCol)
+
+    def setOptimizer(self, value):
+        return self._set(optimizer=value)
+
+    def getOptimizer(self):
+        return self.getOrDefault(self.optimizer)
+
+    def setLoss(self, value):
+        return self._set(loss=value)
+
+    def getLoss(self):
+        return self.getOrDefault(self.loss)
+
+    def _validateParams(self):
+        """
+        Check Param values so we can throw errors on the driver, rather than workers.
+        :return: True if parameters are valid
+        """
+        if not self.isDefined(self.optimizer):
+            raise ValueError("Required Param optimizer must be specified but was not.")
+        if not kmutil.is_valid_optimizer(self.getOptimizer()):
+            raise ValueError("Named optimizer not supported in Keras: %s" % self.getOptimizer())
+        if not self.isDefined(self.loss):
+            raise ValueError("Required Param loss must be specified but was not.")
+        if not kmutil.is_valid_loss_function(self.getLoss()):
+            raise ValueError("Named loss not supported in Keras: %s" % self.getLoss())
+        if not self.isDefined(self.inputCol):
+            raise ValueError("Input column must be defined")
+        if not self.isDefined(self.outputCol):
+            raise ValueError("Output column must be defined")
+        return True
+
+    def _localFit(self, featuresBc, labelsBc, modelBytesBc, paramMap):
+        # Copy the estimator to add to the Params without modifying this instance.
+        estimator = self.copy(paramMap)
+        features = featuresBc.value
+        labels = None if labelsBc is None else labelsBc.value
+        model = kmutil.bytes_to_model(modelBytesBc.value)
+
+        model.compile(optimizer=estimator.getOptimizer(), loss=estimator.getLoss())
+        fit_params = estimator.getKerasFitParams()
+
+        model.fit(x=features, y=labels, **fit_params)
+        return kmutil.model_to_bytes(model)
+
+    def _getNumpyFeaturesAndLabels(self, dataset):
+        image_uri_col = self.getInputCol()
+        label_col = None
+        if self.isDefined(self.labelCol) and self.getLabelCol() != "":
+            label_col = self.getLabelCol()
+        tmp_image_col = self._loadedImageCol()
+        image_df = self.loadImagesInternal(dataset, image_uri_col).dropna(subset=[tmp_image_col])
+
+        localFeatures = []
+        localLabels = []
+        is_with_labels = label_col is not None
+        for row in image_df.collect():
+            spimg = row[tmp_image_col]
+            features = imageStructToArray(spimg)
+            localFeatures.append(features)
+            if is_with_labels:
+                _label = int(row[label_col])
+                localLabels.append(_label)
+
+        if len(localFeatures) == 0:
+            raise ValueError("Given empty dataset!")
+
+        # We must reshape input data to form to the same size so as to be stacked
+        X = np.stack(localFeatures, axis=0)
+        if not is_with_labels:
+            return X, None
+
+        y_category = np.asarray(localLabels).ravel()
+        if self.isDefined(self.labelCardinality):
+            category_field_size = self.getLabelCardinality()
+        else:
+            category_field_size = np.max(y_category) + 1
+            warn_msg = 'inferring label cardinality from training labels {}'
+            warn_msg = warn_msg.format(category_field_size)
+            logger.warning(warn_msg)
+
+        # We also need to convert them to one-hot encoding
+        if self.getIsOneHotLabel():
+            y = y_category
+            assert y.shape[1] == category_field_size, \
+                "label categories must conform {} != {}".format(y.shape[1], category_field_size)
+        else:
+            local_train_set_size = y_category.shape[0]
+            y_one_hot = np.zeros((local_train_set_size, category_field_size))
+            y_one_hot[np.arange(local_train_set_size), y_category] = 1
+            y = y_one_hot
+            info_msg = 'shape of the training labels {}'.format(y.shape)
+            logger.warning(info_msg)
+
+        return X, y
+
+    def _collectModels(self, kerasModelsBytesRDD):
+        """
+        Collect Keras models on workers to MLlib Models on the driver.
+        :param kerasModelBytesRDD: RDD of (param_map, model_bytes) tuples
+        :param paramMaps: list of ParamMaps matching the maps in `kerasModelsRDD`
+        :return: list of MLlib models
+        """
+        transformers = []
+        for (param_map, model_bytes) in kerasModelsBytesRDD.collect():
+            model_filename = kmutil.bytes_to_h5file(model_bytes)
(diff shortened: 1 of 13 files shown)


python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,301 @@
...
+from pyspark.sql.functions import udf
+
+from sparkdl.image.imageIO import imageArrayToStruct, imageStructToArray, imageSchema
+from sparkdl.transformers.param import (
+    keyword_only, HasInputCol, HasOutputCol, SparkDLTypeConverters)
+from sparkdl.transformers.tf_image import TFImageTransformer, OUTPUT_MODES
+import sparkdl.utils.jvmapi as JVMAPI
+import sparkdl.utils.keras_model as keras_utils
+
+if False:
```
phi-dbq: Some version of this will go into test and docs. 
Just put them here to show how it is being developed.

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,301 @@
...
+    import hashlib
+
+    import PIL.Image
+    import requests
+    from io import BytesIO
+    from keras.applications.imagenet_utils import preprocess_input
+
+    def load_from_uri(uri):
+        resp = requests.get(uri)
+        img = PIL.Image.open(BytesIO(resp.content)).convert('RGB')
```
phi-dbq: Probably not the most ideal way to convert RGBA or YCbCr or LA into RGB. 
We could address that later on in `imageIO` improvements.

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,301 @@
...
+        features = featuresBc.value()
+        labels = None if labelsBc is None else labelsBc.value()
+        model = keras_utils.bytes_to_model(modelBytesBc.value())
+
+        model.compile(optimizer=estimator.getOptimizer(), loss=estimator.getLoss())
+        kerasParams = estimator.getKerasParams()
+        model.fit(x=features, y=labels, **kerasParams)
+        return model
+
+    def _getNumpyFeaturesAndLabels(self, dataset):
```
phi-dbq: Features might fail to load due to various resource acquisition errors. 
Thus we must get features and their labels together.

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -31,7 +29,137 @@
     keyword_only, HasInputCol, HasOutputCol, SparkDLTypeConverters)
 from sparkdl.transformers.tf_image import TFImageTransformer, OUTPUT_MODES
 import sparkdl.utils.jvmapi as JVMAPI
-import sparkdl.utils.keras_model as keras_utils
+# Uncomment this for testing 
+#import sparkdl.utils.keras_model as keras_utils
+
+logger = logging.getLogger('sparkdl')
+
+##### ##### ##### ##### ##### ##### ##### ##### ##### ##### ##### #####
```
phi-dbq: These are for local REPL based development ~ they will be moved to utility files.

jkbradley:
reviewing now

python/tests/estimators/test_keras_estimators.py
```diff
@@ -0,0 +1,92 @@
...
+import tempfile
+
+import PIL.Image
+import numpy as np
+import tensorflow as tf
+from keras.applications import Xception
+from keras.applications.imagenet_utils import preprocess_input
+
+from pyspark import SparkContext
+from pyspark.sql import DataFrame, Row
```
jkbradley: eliminate unused imports

python/tests/estimators/test_keras_estimators.py
```diff
@@ -0,0 +1,92 @@
...
+    img_tnsr = preprocess_input(img_arr[np.newaxis, :])
+    return img_tnsr
+
+class KerasEstimatorsTest(SparkDLTestCase):
+
+    def test_valid_params(self):
+        temp_dir = tempfile.mkdtemp()
+        model_filename = os.path.join(temp_dir, 'model.h5')
+
+        estimator = (KerasImageFileEstimator()
```
jkbradley: style: Why is this enclosed in parentheses?

jkbradley: Btw, we generally pass named arguments to constructors in Python, rather than using setters; either is Ok though

python/tests/estimators/test_keras_estimators.py
```diff
@@ -0,0 +1,92 @@
...
+                     .setModelFile(model_filename))
+
+        self.assertTrue(estimator._validateParams())        
+        shutil.rmtree(temp_dir, ignore_errors=True)
+
+    def test_valid_workflow(self):        
+        # Saving a model
+        model = Xception(weights=None)
+        temp_dir = tempfile.mkdtemp()
+        model_filename = os.path.join(temp_dir, 'model.h5')
```
jkbradley: Since this path is expected in the other test, how about putting this set up in outer setUpClass/tearDownClass methods (but be sure to call the parent setup/teardown methods).

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,305 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
phi-dbq: Got it

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,305 @@
...
+from sparkdl.transformers.param import (
+    keyword_only, HasInputCol, HasOutputCol, SparkDLTypeConverters)
+from sparkdl.transformers.keras_image import KerasImageFileTransformer
+from sparkdl.transformers.tf_image import TFImageTransformer, OUTPUT_MODES
+import sparkdl.utils.jvmapi as JVMAPI
+from sparkdl.utils.keras_model import *
+
+logger = logging.getLogger('sparkdl')
+
+class CanLoadImage(Params):
```
phi-dbq: `KerasImageFileTransformer` and `TFImageTransformer` require a trained Keras/TensorFlow model. 
Here though we only do image decoding (which might not actually use TensorFlow). 
I am thinking about using it as a common ABC and let the image transformers/estimators extend it.

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,305 @@
...
+                        "with shape (batch_size (1), height, width, num_channels).")
+
+    def setImageLoader(self, value):
+        return self._set(imageLoader=value)
+
+    def getImageLoader(self):
+        return self.getOrDefault(self.imageLoader)
+
+    def _loadedImageCol(self):
+        return "__sdl_img"
```
phi-dbq: Indeed, it might be better to have some randomness in the name to avoid potential conflict.

MrBago:
@phi-dbq Sorry I've been putting this review off. What's the current status? Are you looking for a review of the APIs and design or are you ready for a full pass?

phi-dbq:
@MrBago Thank you very much! I updated the PR description. Would you like to take a look at the first two parts (MLlib Params and Keras utilities)? Then we can proceed with the Keras Estimator.

python/sparkdl/param/image_params.py
```diff
@@ -0,0 +1,93 @@
...
+    def setImageLoader(self, value):
+        return self._set(imageLoader=value)
+
+    def getImageLoader(self):
+        return self.getOrDefault(self.imageLoader)
+
+    def _loadedImageCol(self):
+        return "__sdl_img"
+    
+    def _loadImages(self, dataset, inputCol):
```
thunterdb: this is not really private but protected. I would rather name it without a `_` and mention in the doc that is internal, or call it `loadImagesInternal`

python/sparkdl/param/image_params.py
```diff
@@ -0,0 +1,93 @@
...
+        return self.getOrDefault(self.modelFile)
+
+    def setKerasParams(self, value):
+        return self._set(kerasParams=value)
+    
+    def getKerasParams(self):
+        return self.getOrDefault(self.kerasParams)
+
+
+class CanLoadImage(Params):
```
thunterdb: this trait is not useful as it stands for users, as the implementation of the loader `_load_image_from_uri` is only in the test. How do you see users using it? We have to provide a default value for images.

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,262 @@
...
+
+        return transformers
+
+    def _fitInParallel(self, dataset, paramMaps):
+        """
+        Fits len(paramMaps) models in parallel, one in each Spark task.
+        :param paramMaps: non-empty list or tuple of ParamMaps (dict values)
+        :return: list of fitted models, matching the order of paramMaps
+        """
+        #sc = dataset.sql_ctx._sc
```
thunterdb: unused

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,262 @@
...
+        print('DataSet in _fitInParallel')
+        dataset.printSchema()
```
thunterdb: remove

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,262 @@
...
+    def fit(self, dataset, params=None):
+        """
+        Fits a model to the input dataset with optional parameters.
+        :param dataset: input dataset, which is an instance of :py:class:`pyspark.sql.DataFrame`.
+                        The column `inputCol` should be of type
+                        `sparkdl.image.imageIO.imgSchema`.
+        :param params: An optional param map that overrides embedded params. If a list/tuple of
+                       param maps is given, this calls fit on each param map and returns a list of
+                       models.
+        :returns: fitted model(s).  If params includes a list of param maps, the order of these
```
thunterdb: This API is not great because it returns a list of all the evaluated models. If you look at the other meta estimators in MLlib, they return the best one. Since the model evaluation is currently embedded in Keras (and outside the reach of Spark for now), this compromise is acceptable, but we will need to figure out how to make it return the best model eventually. Can you put a comment saying that this return value is subject to change in the future?

cc @jkbradley who may have some thoughts on this point.

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,262 @@
...
+        return model
+
+    def _getNumpyFeaturesAndLabels(self, dataset):
+        image_uri_col = self.getInputCol()
+        label_col = None
+        if self.isDefined(self.labelCol) and self.getLabelCol() != "":
+            label_col = self.getLabelCol()
+        tmp_image_col = self._loadedImageCol()
+        image_df = self._loadImages(dataset, image_uri_col).dropna(subset=[tmp_image_col])
+        localFeatures = list(map(
```
thunterdb: This is an impressive 2-liner, but please use list comprehensions for readability. Note that you can distribute the array construct. It does not matter too much though as this is something we are probably going to move to the scala side.

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,262 @@
...
+        """
+        setParams(self, inputCol=None, outputCol=None, labelCol=None, modelFile=None,
+                  imageLoader=None, outputMode="vector", optimizer=None, loss=None,
+                  kerasParams=None):
+        """
+        kwargs = self._input_kwargs
+        self._set(**kwargs)
+        return self
+
+    def setLabelCol(self, value):
```
thunterdb: for usability, we should have a `setModel` that takes a Keras model and does all the file writing. This is how most users will want to interact with it. Can you add a TODO for the future?

python/tests/estimators/test_keras_estimators.py
```diff
@@ -0,0 +1,75 @@
...
+    img_tnsr = preprocess_input(img_arr[np.newaxis, :])
+    return img_tnsr
+
+class KerasEstimatorsTest(SparkDLTestCase):
+
+    def setUp(self):
+        self.temp_dir = tempfile.mkdtemp()
+        self.model_filename = os.path.join(self.temp_dir, 'model.h5')
+
+        self.estimator = KerasImageFileEstimator(inputCol='imageUri',
```
phi-dbq: @MrBago any thoughts?

python/sparkdl/param/image_params.py
phi-dbq: The user will provide an image loading function. This has been the assumption throughout the project.

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,263 @@
...
+
+        localLabels = []
+        for row in image_df.select(label_col).rdd.toLocalIterator():
+            localLabels.append(int(row.label))
+        if len(localLabels) != len(localFeatures):
+            raise ValueError("Number of labels {} not matching number of instances {}"
+                             .format(len(localLabels), len(localFeatures)))
+        # We also need to convert them to one-hot encoding
+        y_category = np.stack(localLabels, axis=0)
+        batch_size = y_category.shape[0]
```
MrBago: The use of `stack` and `shape` is confusing here because it makes it seem like `y_category` is multidimensional. Could we do something like:

```
localLabel = np.asarray(localLabel)
batch_size = len(localLabel)
```

python/tests/estimators/test_keras_estimators.py
```diff
@@ -0,0 +1,81 @@
...
+    def tearDown(self):
+        shutil.rmtree(self.temp_dir, ignore_errors=True)
+
+    def test_valid_params(self):
+        self.assertTrue(self.estimator._validateParams())
+
+    def test_valid_workflow(self):
+        # Saving a model
+        model = Xception(weights=None)
+        model.save(self.model_filename)
```
MrBago: It's a bit weird to build the estimator and `model_filename` and setUp but not write out the actual model until we're inside the test. If other tests wanted to use this estimator, they'd have to delete this file and replace it.

Why don't we put the estimator creation inside the test for now, we can promote it to setup if we write other tests that need it.

MrBago:
I just saw that some of my comments were on an older version of this PR, sorry about that I'm not sure how that happened.

python/tests/estimators/test_keras_estimators.py
```diff
@@ -0,0 +1,75 @@
...
+
+        # Create image URI dataframe
+        image_uris = getSampleImagePaths()
+        image_labels = [1] * len(image_uris)
+        image_uri_df = self.sc.parallelize(
+            zip(image_uris, image_labels)).toDF(['imageUri', 'label'])
+        
+        transformers = self.estimator.fit(image_uri_df)
+        self.assertTrue(1 == len(transformers))
```
MrBago: `self.assertEqual` is a little more informative than `assertTrue` when it fails.

python/sparkdl/param/image_params.py
```diff
@@ -0,0 +1,172 @@
...
+        return self.getOrDefault(self.outputMode)
+
+
+class HasLabelCol(Params):
+
+    labelCol = Param(Params._dummy(), "labelCol",
+                     "name of the column storing the training data labels",
+                     typeConverter=TypeConverters.toString)
+    
+    labelCardinality = Param(Params._dummy(), "labelCardinality",
```
phi-dbq: The `label` column is expected to be numerical. The training data might not have all the label presented, so we also opt for users to provide a `cardinarlity` parameter. Most Keras models expect one-hot encoded labels. If the input labels are not in that format, we convert it.

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,263 @@
...
+        y_one_hot = np.zeros((batch_size, 1000))
```
phi-dbq: It was the default imagenet number of class labels. Removed it.

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,305 @@
...
+        image_df = self._loadImages(dataset, image_uri_col).dropna(subset=[tmp_image_col])
+        localFeatures = list(map(
+            lambda row: imageStructToArray(row.asDict()[tmp_image_col]),
+            image_df.select(tmp_image_col).collect()))
+        if len(localFeatures) == 0:
+            raise ValueError("Given empty dataset!")        
+
+        # We must reshape input data to form to the same size so as to be stacked
+        print('Warning: user provided imageLoader must convert all images to the same size')
+        X = np.stack(localFeatures, axis=0)
```
phi-dbq: It looks like `numpy.stack` takes care of that. I thus removed the warning.

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,305 @@
...
+
+        if label_col:
+            print('[_getNumpyFeaturesAndLabels] label_col', label_col)
```
phi-dbq: Removed that, too

phi-dbq:
Could probably add more documentation in another PR.

MrBago:
I had a quick pass over this and I'm a little confused. I wanted to ask a few questions before doing a full pass. It seems like you've added 2 params to "KerasImageFileEstimator", `labelCardinality`, and `isOneHotLabel`.

I was surprised to see these params on the Estimator and not on the transformer. What does the transformer do when these params are set/not set. Do we actually need these to be params, can't the transformer infer this based on the data type (vertors/arrays are treated as one-hot and integers are treated as class values, cardinality can be computed from the data during training).

python/sparkdl/estimators/keras_image_file_estimator.py
phi-dbq: Discussed offline with @jkbradley and @thunterdb.

phi-dbq:
@MrBago @thunterdb would you like to give it another round? Thank you!

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,262 @@
...
+            y_category = np.stack(localLabels, axis=0)
+            batch_size = y_category.shape[0]
+            y_one_hot = np.zeros((batch_size, 1000))
+            y_one_hot[np.arange(batch_size), y_category] = 1
+            y = y_one_hot
+            print('[_getNumpyFeaturesAndLabels] label one-hot', y.shape)
+        else:
+            y = None
+            
+        print('[_getNumpyFeaturesAndLabels] features', X.shape)
```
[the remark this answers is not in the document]
phi-dbq: done

python/sparkdl/param/image_params.py
```diff
@@ -0,0 +1,190 @@
...
+OUTPUT_MODES = ["vector", "image"]
+
+class HasInputImageNodeName(Params):
+
+    inputImageNodeName = Param(Params._dummy(), "inputImageNodeName",
+                               "name of the graph element/node corresponding to the input",
+                               typeConverter=TypeConverters.toString)
+
+    def __init__(self):
+        super(HasInputImageNodeName, self).__init__()
```
MrBago: I don't think you need this. A class will inherit its parent's methods.

python/sparkdl/param/image_params.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+class HasOutputNodeName(Params):
+
+    outputNodeName = Param(Params._dummy(), "outputImageNodeName",
+                           "name of the graph element/node corresponding to the output",
+                           typeConverter=TypeConverters.toString)
+
+    def __init__(self):
+        super(HasOutputNodeName, self).__init__()
```
MrBago: Same as above.

python/sparkdl/param/image_params.py
```diff
@@ -0,0 +1,190 @@
...
+class HasOutputMode(Params):
+
+    outputMode = Param(Params._dummy(), "outputMode",
+                       "How the output column should be formatted. 'vector' for a 1-d MLlib " +
+                       "Vector of floats. 'image' to format the output to work with the image " +
+                       "tools in this package.",
+                       typeConverter=SparkDLTypeConverters.supportedNameConverter(OUTPUT_MODES))
+
+    def __init__(self):
+        super(HasOutputMode, self).__init__()
```
MrBago: same

python/sparkdl/param/image_params.py
```diff
@@ -0,0 +1,190 @@
...
+    .. note:: The Estimator expect this columnd to contain data directly usable for the Keras model.
+              This usually means that the labels are already encoded in one-hot format.
+              Please consider adding a :py:obj:`OneHotEncoder` to transform the label column.
+    """
+    labelCol = Param(Params._dummy(), "labelCol",
+                     "name of the column storing the training data labels",
+                     typeConverter=TypeConverters.toString)
+
+    def __init__(self):
+        super(HasLabelCol, self).__init__()
```
MrBago: same

python/sparkdl/param/image_params.py
phi-dbq: Haven't thought much about them as the MLlib params are created from a script (and these are created along the way). I will remove them.

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,267 @@
...
+        localFeaturesBc = sc.broadcast(localFeatures)
+        localLabelsBc = None if localLabels is None else sc.broadcast(localLabels)
+
+        # Broadcast Keras model (HDF5) file content as bytes
+        modelBytes = self._loadModelAsBytes()
+        modelBytesBc = sc.broadcast(modelBytes)
+
+        kerasModelBytesRDD = paramMapsRDD.map(
+            lambda paramMap: (
+                paramMap, self._localFit(localFeaturesBc, localLabelsBc, modelBytesBc, paramMap)))
```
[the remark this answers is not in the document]
phi-dbq: That's a good question. Each individual task will create a `Transformer` which copies the entire set of parameters from `self`.

python/tests/transformers/keras_image_test.py
```diff
@@ -28,13 +28,12 @@ def test_loadImages(self):
         output_col = "preds"
 
         model_path = image_utils.prepInceptionV3KerasModelFile("inceptionV3.h5")
-        transformer = KerasImageFileTransformer(inputCol=input_col, outputCol=output_col,
```
phi-dbq: changed: line width >= 100

python/tests/transformers/keras_image_test.py
```diff
@@ -51,10 +50,9 @@ def test_inceptionV3_vs_keras(self):
...
```
phi-dbq: changed: line width >= 100

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,284 @@
...
+                    "Please contact library mantainers to file a bug"]
+        raise NotImplementedError('\n'.join(err_msgs))
+
+    def _validateFitParams(self, params):
+        """ Check if an input parameter set is valid """
+        if isinstance(params, (list, tuple, dict)):
+            assert self.getInputCol() not in params, \
+                "params {} cannot contain input column {}".format(params, self.getInputCol())
+
+    def fit(self, dataset, params=None):
```
thunterdb: Nit: move this function before all the private ones, to put all the public pictures together.

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,284 @@
...
+            raise ValueError("Given empty dataset!")
+
+        # We must reshape input data to form to the same size so as to be stacked
+        X = np.stack(localFeatures, axis=0)
+        y = None
+        if is_with_labels:
+            if not localLabels:
+                raise ValueError("Failed to load any labels from dataset, but labels are required")
+            y = np.stack(localLabels, axis=0)
+            assert y.shape[0] == X.shape[0], "must have same amount of features and labels"
```
thunterdb: also add the shapes in the string to give debug info

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,284 @@
...
+        image_uri_col = self.getInputCol()
+        label_col = None
+        if self.isDefined(self.labelCol) and self.getLabelCol() != "":
+            label_col = self.getLabelCol()
+        tmp_image_col = self._loadedImageCol()
+        image_df = self.loadImagesInternal(dataset, image_uri_col).dropna(subset=[tmp_image_col])
+
+        localFeatures = []
+        localLabels = []
+        is_with_labels = label_col is not None
```
thunterdb: this will be cleaner if you do it this way:

```
rows = image_df.collect()
... build X ...
if label_col is not None:
   ... build y ...
```

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,284 @@
...
+
+    def _collectModels(self, kerasModelsBytesRDD):
+        """
+        Collect Keras models on workers to MLlib Models on the driver.
+        :param kerasModelBytesRDD: RDD of (param_map, model_bytes) tuples
+        :param paramMaps: list of ParamMaps matching the maps in `kerasModelsRDD`
+        :return: list of MLlib models
+        """
+        transformers = []
+        for (param_map, model_bytes) in kerasModelsBytesRDD.collect():
```
thunterdb: that may totally blow up, but I guess this is ok for now...

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,284 @@
...
+        # Broadcast Keras model (HDF5) file content as bytes
+        modelBytes = self._loadModelAsBytes()
+        modelBytesBc = sc.broadcast(modelBytes)
+
+        # Obtain params for this estimator instance
+        baseParamMap = self.extractParamMap()
+        baseParamDict = dict([(param.name, val) for param, val in baseParamMap.items()])
+        baseParamDictBc = sc.broadcast(baseParamDict)
+
+        # Create useful parameters from paramMap
```
thunterdb: this comment does not correspond to what the function is doing

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,284 @@
...
+        Load the Keras model file as a byte string.
+        :return: str containing the model data
+        """
+        with open(self.getModelFile(), mode='rb') as fin:
+            fileContent = fin.read()
+        return fileContent
+
+    def _fit(self, dataset):  # pylint: disable=unused-argument
+        err_msgs = ["This function should not have been called",
+                    "Please contact library mantainers to file a bug"]
```
thunterdb: `maintainers`

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,284 @@
...
+        raise NotImplementedError('\n'.join(err_msgs))
+
+    def _validateFitParams(self, params):
+        """ Check if an input parameter set is valid """
+        if isinstance(params, (list, tuple, dict)):
```
thunterdb: else?

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,284 @@
...
+            labels = None if localLabelsBc is None else localLabelsBc.value
+            _fit_params = params['kerasFitParams']
+            model.fit(x=features, y=labels, **_fit_params)
+
+            return kmutil.model_to_bytes(model)
+
+        kerasModelBytesRDD = paramMapsRDD.map(lambda paramMap: (paramMap, _local_fit(paramMap)))
+        return self._collectModels(kerasModelBytesRDD)
+
+    def _loadModelAsBytes(self):
```
thunterdb: we will need to change `HasKerasModel` eventually: storing the name of a file is not enough for import/export. No need to change in this PR though.

python/sparkdl/estimators/keras_image_file_estimator.py
phi-dbq: Added warning in `.fit` function (docs are only shown in public functions).

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,267 @@
...
+        Check Param values so we can throw errors on the driver, rather than workers.
+        :return: True if parameters are valid
+        """
+        if not self.isDefined(self.optimizer):
+            raise ValueError("Required Param optimizer must be specified but was not.")
+        if not kmutil.is_valid_optimizer(self.getOptimizer()):
+            raise ValueError("Named optimizer not supported in Keras: %s" % self.getOptimizer())
+        if not self.isDefined(self.loss):
+            raise ValueError("Required Param loss must be specified but was not.")
+        if not kmutil.is_valid_loss_function(self.getLoss()):
```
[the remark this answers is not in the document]
phi-dbq: moved them to type converters

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,267 @@
...
+        return self.getOrDefault(self.loss)
+
+    def _validateParams(self):
+        """
+        Check Param values so we can throw errors on the driver, rather than workers.
+        :return: True if parameters are valid
+        """
+        if not self.isDefined(self.optimizer):
+            raise ValueError("Required Param optimizer must be specified but was not.")
+        if not kmutil.is_valid_optimizer(self.getOptimizer()):
```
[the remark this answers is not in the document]
phi-dbq: byte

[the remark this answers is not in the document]
phi-dbq: moved to type convertes

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,267 @@
...
+
+    # TODO(phi-dbq): allow users to pass a Keras optimizer object
+    #                so that they can configure learning rate, momentum, etc.
+    optimizer = Param(Params._dummy(), "optimizer",
+                      "named Keras optimizer (str), please refer to https://keras.io/optimizers",
+                      typeConverter=TypeConverters.toString)
+
+    loss = Param(Params._dummy(), "loss",
+                 "named Keras loss (str), please refer to https://keras.io/losses",
+                 typeConverter=TypeConverters.toString)
```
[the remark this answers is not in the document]
phi-dbq: moved to type converters

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -0,0 +1,317 @@
...
+        """
+        kwargs = self._input_kwargs
+        return self._set(**kwargs)
+
+    def fit(self, dataset, params=None):
+        """
+        Fits a model to the input dataset with optional parameters.
+
+        .. warning:: This returns the byte serialized HDF5 file for each model to the driver.
+                     If the model file is way too large, the driver might go out-of-memory.
```
thunterdb: ~way too~

thunterdb:
Docs look very good, thank you Philip. @sueann , do you have some more comments?

jkbradley:
Given @thunterdb 's approval, I'll merge this

allwefantasy:
“We assume the training data fits in memory on a single server. The input dataframe is converted to numerical image features and broadcast to all the worker nodes.”  Is there any way we can stream data to  every executor?  Driver will suffer memory issue cause we first collect trainning data and model in driver then broadcast

PR #39, opened 2017-08-08
[API] TensorFlow Graph Transformer

phi-dbq:

API design and reference implementation of a Spark MLlib Transformer from any TensorFlow Graph.

**Notice** 
This implementation is only a proof-of-concept reference. We will break it up into multiple PRs with finer grain tests and documentations. 

This is not an implementation of the (attention-based) Transformer architecture (https://arxiv.org/abs/1706.03762). 

## How is this patch tested?

- [x] Integration tests
- [x] Manual tests


python/sparkdl/transformers/tf_tensor.py
```diff
@@ -0,0 +1,93 @@
...
+    We assume that 
+    - All graphs have a "minibatch" dimension (i.e. an unknown leading
+      dimension) in the tensor shapes.
+    - Input DataFrame has an array column where all elements have the same length
+
+    .. note:: The input tensorflow graph should have appropriate weights constantified,
+              since a new session is created inside this transformer.
+    """
+
+    graphFunction = Param(Params._dummy(), "graphFunction",
```
phi-dbq: The intention is to provide this as the Transformer API and provide utilities to convert any TensorFlow/Keras graph into a `GraphFunction` object.

sueann: `GraphFunction` and `IsolatedSession` were added as internal classes not to be  exposed to users, only to be used within the internals of sparkdl, so we shouldn't have `GraphFunction` be a param user defines. These are foreign concepts to TensorFlow users. We should stick with what users know as much as possible so that this functionality is as easy for them to use as possible. Making this API match the one TFoS guys are building (https://github.com/yahoo/TensorFlowOnSpark/pull/114/files) would be better. If you think the proposed API in https://github.com/yahoo/TensorFlowOnSpark/pull/114/files is not as good, we should discuss.

python/sparkdl/transformers/tf_tensor.py
```diff
@@ -0,0 +1,93 @@
...
+            feeds, fetches = issn.importGraphFunction(gfn, prefix='')
+            assert len(feeds) == 1, 'only support single input TF model'
+            assert len(fetches) == 1, 'only support single output TF model'
+
+            orig_in_name = tfx.op_name(issn.graph, feeds[0])
+            input_df = analyzed_df.withColumnRenamed(
+                self.getInputCol(), orig_in_name).select(orig_in_name)
+            output_df = tfs.map_blocks(fetches, input_df)
+            orig_out_name = tfx.op_name(issn.graph, fetches[0])
+            final_df = output_df.withColumn(
```
sueann: you can do a rename for the column instead

python/sparkdl/transformers/tf_tensor.py
sueann: We may not need all the params the other API has, but if that is the case, we should have this API start with params that are a strict subset of the other API.

phi-dbq:
I am designing the API around TensorFlow Graph + Input/Output Tensors.
The TFoS API takes a checkpoint directory for the (meta) graph. We can get a utility function to convert it to a TensorFlow Graph.

python/sparkdl/transformers/param.py
```diff
@@ -91,12 +94,23 @@ def getOutputCol(self):
         """
         return self.getOrDefault(self.outputCol)
 
-
 # New in sparkdl
 
 class SparkDLTypeConverters(object):
 
     @staticmethod
+    def toTFGraph(value):
```
phi-dbq: Directly importing `tf.GraphDef` without the input tensor mapping is not working.
Thus the user either provides a `TensorFlow Graph` or provide a `GraphFunction` (for developers).

python/sparkdl/transformers/param.py
```diff
@@ -45,6 +46,8 @@ def wrapper(self, *args, **kwargs):
     return wrapper
 
 
+""" TensorFlow Specific Parameters """
```
sueann: hasInputCol is not TensorFlow specific. in fact, it was copied from pyspark...!

python/sparkdl/transformers/tf_tensor.py
```diff
@@ -0,0 +1,105 @@
...
+    """ Apply a transform with a graph function
+    """
+    analyzed_df = tfs.analyze(df)
+
+    with IsolatedSession() as issn:
+        feeds, fetches = issn.importGraphFunction(gfn, prefix='')
+        assert len(feeds) == 1, 'only support single input TF model'
+        assert len(fetches) == 1, 'only support single output TF model'
+
+        # Change the column name for TensorFrames
```
sueann: it's not much code -- why not just put it in the transformer and skip having to convert from/to GraphFunction? also can keep using the IsolatedSession there then.

python/sparkdl/transformers/tf_tensor.py
```diff
@@ -0,0 +1,105 @@
...
+
+        # Change the names back
+        orig_out_name = tfx.op_name(issn.graph, fetches[0])
+        final_df = (output_df
+                    .withColumnRenamed(orig_out_name, output_col)
+                    .withColumnRenamed(orig_in_name, input_col))
+
+    return final_df
+
+class TFOneDimTensorTransformer(Transformer, HasInputCol, HasOutputCol,
```
sueann: I'd just name it TFTransformer or TFModel. We can add to the doc that it only supports 1-d arrays currently.

python/tests/transformers/tf_tensor_test.py
```diff
@@ -0,0 +1,157 @@
...
+            transfomer = TFOneDimTensorTransformer(
+                tfGraph=sess.graph, inputTensor=x, outputTensor=z,
+                inputCol='vec', outputCol='outCol')
+            final_df = transfomer.transform(analyzed_df)
+            out_tgt = grab_df_arr(final_df, 'outCol')
+            
+        self.assertTrue(np.allclose(out_ref, out_tgt))
+            
+
+    def test_simple_graph_function(self):
```
sueann: what does this test? it seems like the difference between this and test_simple() is that this one uses GraphFunction and IsolatedSession? I don't think that test belongs in this class (and the functionality should have been tested in the Suites for GraphFunction/IsolatedSession anyway).

python/sparkdl/transformers/param.py
```diff
@@ -122,3 +134,49 @@ def converter(value):
...
+        super(HasOutputTensor, self).__init__()
+
+    def setOutputTensor(self, value):
+        return self._set(outputTensor=value)
+
+    def getOutputTensor(self):
+        return self.getOrDefault(self.outputTensor)
+
+
+class HasInputTensor(Params):
```
anfeng: Normally, we need to handle multiple 1+ input tensors. See TensorFlow Serving as a reference.

python/sparkdl/transformers/tf_tensor.py
```diff
@@ -0,0 +1,105 @@
...
+        __init__(self, tfGraph=None, inputTensor=None, outputTensor=None,
+                 inputCol=None, outputCol=None)
+        """
+        super(TFOneDimTensorTransformer, self).__init__()
+        kwargs = self._input_kwargs
+        self.setParams(**kwargs)
+
+    @keyword_only
+    def setParams(self, tfGraph=None, inputTensor=None, outputTensor=None,
+                  inputCol=None, outputCol=None):
```
anfeng: We need to handle multiple input colmns.

anfeng: Also for output columns

python/sparkdl/transformers/param.py
anfeng: Also for output tensors. See https://github.com/tensorflow/serving/blob/master/tensorflow_serving/example/mnist_saved_model.py#L109-L114

python/sparkdl/transformers/tf_tensor.py
phi-dbq: I am wondering if we want the tensors to be in separate columns or in fields of a nested column?

python/sparkdl/transformers/tf_tensor.py
```diff
@@ -0,0 +1,94 @@
...
+import sparkdl.graph.utils as tfx
+from sparkdl.transformers.param import (
+    keyword_only, HasInputMapping, HasOutputMapping, SparkDLTypeConverters,
+    HasTFGraph, HasTFHParams)
+
+__all__ = ['TFTensorTransformer']
+
+logger = logging.getLogger('sparkdl')
+
+class TFTensorTransformer(Transformer, HasTFGraph, HasTFHParams, HasInputMapping, HasOutputMapping):
```
sueann: TFTensorTransformer is redundant since it spells out to be TensorFlowTensorTransformer. Why not just TFTransformer or TFModel?

python/sparkdl/transformers/tf_tensor.py
```diff
@@ -69,37 +42,53 @@ class TFOneDimTensorTransformer(Transformer, HasInputCol, HasOutputCol,
...
+                fetches.append(tfx.get_tensor(issn.graph, tnsr))
+                old_colname = tfx.op_name(issn.graph, tnsr)
+                output_renaming[old_colname] = new_colname
+
+            gfn = issn.asGraphFunction(feeds, fetches)
+
+        analyzed_df = tfs.analyze(df)
+
+        with IsolatedSession() as issn:
+            _, fetches = issn.importGraphFunction(gfn, prefix='')
```
sueann: Lines 83-33: why do you need to create a GraphFunction and then importGraphFunction to get fetches vs just using the fetches from above? 

Somewhat related: `analyzed_df = tfs.analyze(df)` (line 85) doesn't have to be called right there, right? Could be called before the first `IsolatedSession` here?

phi-dbq: The use of two sessions is due to the fact that exporting to `GraphFunction` modifies the graph and it has to be imported in another session.

sueann: oh i see... if we just need strip_and_freeze we should just call that here directly because this is hiding the fact that that is what (and all) we are doing.

phi-dbq: It also seems that importing graph_def directly is a bit flaky. `importGraphFunction` attaches the names and tensors correctly. It might just save us a few lines of code.

python/tests/transformers/tf_tensor_test.py
```diff
@@ -0,0 +1,184 @@
...
+            # Rename the output column (by default, the name of the fetch op's name)
+            orig_out_name = tfx.op_name(issn.graph, fetches[0])
+            final_df = output_df.withColumnRenamed(orig_out_name, output_col)
+
+        arr_ref = grab_df_arr(final_df, output_col)
+
+        # Using the Transformer
+        transformer = TFModelTransformer(tfGraph=gfn,
+                                         inputMapping={
+                                             input_col: gfn.input_names[0]
```
phi-dbq: We can either provide names or tensors for the mappings

python/tests/transformers/tf_tensor_test.py
```diff
@@ -38,6 +41,55 @@ def _get_rand_vec_df(self, num_rows, vec_size):
...
+            for row in df.rdd.toLocalIterator():
+                arr = np.array(row.vec)[np.newaxis, :]
+                _results.append(sess.run(z, {x: arr}))
+            out_ref = np.hstack(_results)
+
+        # Load the saved model checkpoint
+        # We want to clear device assignment in order to run it anywhere we want
+        with IsolatedSession() as issn:          
+            saver = tf.train.import_meta_graph('{}.meta'.format(saved_path), clear_devices=True)
+            saver.restore(issn.sess, saved_path)
```
phi-dbq: We have to export this checkpoint as `GraphFunction` in the same session where it is restored. Since the variables are only initialized in this session.

phi-dbq: Also it seems that we have to import the `meta_graph` first or there won't be any graph in this session (tried with either `IsolatedSession` or standard 'tf.Session').

python/tests/transformers/tf_tensor_test.py
```diff
@@ -0,0 +1,236 @@
...
+        model_temp_dir = tempfile.mkdtemp()        
+        ckpt_dir = os.path.join(model_temp_dir, 'model_ckpt')
+        with tf.Session() as sess:
+            x = tf.placeholder(tf.float64, shape=[None, vec_size], name='tnsrIn')
+            w = tf.Variable(tf.random_normal([vec_size], dtype=tf.float64),
+                            dtype=tf.float64, name='varW')
+            z = tf.reduce_mean(x * w, axis=1, name='tnsrOut')            
+            sess.run(w.initializer)
+            saver = tf.train.Saver(var_list=[w])
+            saved_path = saver.save(sess, ckpt_dir, global_step=2702)
```
phi-dbq: Save the model and record the current `global_step`. Then we'd expect files like these.
```
model.ckpt-2702.index
model.ckpt-2702.meta
model.ckpt-2702.data-00000-of-00001
```

python/tests/transformers/tf_tensor_test.py
```diff
@@ -0,0 +1,236 @@
...
+        # We want to clear device assignment in order to run it anywhere we want
+        with IsolatedSession() as issn:          
+            saver = tf.train.import_meta_graph('{}.meta'.format(saved_path), clear_devices=True)
+            saver.restore(issn.sess, saved_path)
+            gfn = issn.asGraphFunction(
+                [tfx.get_tensor(issn.graph, 'tnsrIn')],
+                [tfx.get_tensor(issn.graph, 'tnsrOut')])
+
+        transformer = TFModelTransformer(tfGraph=gfn,
+                                         inputMapping={
```
phi-dbq: We should be better reversing the input mapping to match the `feed_dict` method.

python/sparkdl/graph/utils.py
```diff
@@ -95,31 +95,49 @@ def get_tensor(graph, tfobj_or_name):
...
+        # WARNING: this depends on TensorFlow's tensor naming convention
+        name = tfobj_or_name
+        name_parts = name.split(":")
+        assert len(name_parts) <= 2, name_parts
+        if len(name_parts) < 2:
+            name += ":0"
+        return name
+    elif hasattr(tfobj_or_name, 'graph'):
+        tfobj = tfobj_or_name
+        return get_tensor(tfobj.graph, tfobj).name
```
thunterdb: `tensor_name`

python/sparkdl/graph/utils.py
```diff
@@ -95,31 +95,49 @@ def get_tensor(graph, tfobj_or_name):
...
+    else:
+        raise TypeError('invalid tf.Tensor name query type {}'.format(type(tfobj_or_name)))
 
-def as_op_name(name):
+def as_op_name(tfobj_or_name):
```
thunterdb: We have 5 functions that compute some names in this file, this is at least two too many.

python/sparkdl/transformers/tf_tensor.py
```diff
@@ -0,0 +1,94 @@
...
+
+            fetches = []
+            for tnsr, output_colname in self.getOutputMapping():
+                fetches.append(tfx.get_tensor(issn.graph, tnsr))
+                tf_expected_colname = tfx.op_name(issn.graph, tnsr)
+                output_renaming[tf_expected_colname] = output_colname
+
+            gfn = issn.asGraphFunction(feeds, fetches, strip_and_freeze=True)
+
+        analyzed_df = tfs.analyze(df)
```
thunterdb: The analyze() function is not smart enough currently to know that the data is already analyzed, so it make take a substantial amount of time. This is an easy fix in tensorframes though.

python/sparkdl/transformers/tf_tensor.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+        super(TFTransformer, self).__init__()
+        kwargs = self._input_kwargs
+        return self._set(**kwargs)
+
+    def _transform(self, dataset):
+        df = dataset
+        output_renaming = {}
+
+        with IsolatedSession(graph=self.getTFGraph()) as issn:
```
thunterdb: I see why you need this currently. There is a `feed_dict` arguments for `map_rows` that does exactly what you want in tensorframes. It still needs to be implemented for `map_blocks`, but this is a quick fix.

thunterdb: Once this argument is there, this block of code will not be required anymore.

thunterdb: Also, errors here are going to be confusing because you are renaming the columns.

python/sparkdl/transformers/tf_tensor.py
thunterdb: @phi-dbq what do you mean by flaky? I do not understand why simply using `import_graph_def` would not work here.

python/tests/transformers/tf_tensor_test.py
```diff
@@ -0,0 +1,236 @@
...
+            _results = []
+            for row in df.rdd.toLocalIterator():
+                arr = np.array(row.vec)[np.newaxis, :]
+                _results.append(sess.run(z, {x: arr}))
+            out_ref = np.hstack(_results)
+
+        # Apply the transform
+        transfomer = TFTransformer(tfGraph=graph,
+                                   inputMapping={
+                                       'vec': x
```
thunterdb: why are there new lines here?

python/sparkdl/transformers/tf_tensor.py
phi-dbq: Sometimes one might use a session without resetting the default graph. 
```
with tf.Session as sess:
    tf.import_graph_def(graph_def, name='')
    sess.run(...)
``` 

In case there is a default graph and there is a name conflict, the tensors in the imported graph will be renamed automatically.

python/sparkdl/transformers/param.py
```diff
@@ -107,13 +196,11 @@ def toStringOrTFTensor(value):
                 raise TypeError("Could not convert %s to tensorflow.Tensor or str" % type(value))
 
     @staticmethod
-    def toTFGraph(value):
-        # TODO: we may want to support tf.GraphDef in the future instead of tf.Graph since user
-        # is less likely to mess up using GraphDef vs Graph (e.g. constants vs variables).
-        if isinstance(value, tf.Graph):
+    def toGraphFunction(value):
```
sueann: this is unused?

python/sparkdl/transformers/param.py
```diff
@@ -92,11 +123,69 @@ def getOutputCol(self):
         return self.getOrDefault(self.outputCol)
 
 
-# New in sparkdl
+class HasOutputCols(Params):
```
sueann: unused?

python/sparkdl/transformers/param.py
```diff
@@ -68,12 +74,37 @@ def getInputCol(self):
         return self.getOrDefault(self.inputCol)
 
 
+class HasInputCols(Params):
```
sueann: unused

python/sparkdl/transformers/tf_tensor.py
sueann: I thought the internal logic here might change given the changes in tensorframes, is that not true?

python/sparkdl/transformers/tf_tensor.py
sueann: But in this case there isn't another graph in the session right? So there is no need to have the complex logic around the graph / graphdef. Could lines 70-92 just be:
```
analyzed_df = tfs.analyze(df)

# prune the graph for tensorframes
with tf.session(graph=self.getTFGraph()) as sess:
  fetches = [tfx.get_tensor(sess.graph, tnsr) for tnsr, output_colname in self.getOutputMapping()]
  gdef = tfx.strip_and_freeze_until(fetches, sess.graph, sess)

with tf.session() as sess2:
  output_names = [tfx.validated_output(sess.graph, elem) for elem in fetches]
  tf.import_graph_def(gdef,
                                    input_map=None,
                                    return_elements=output_names,
                                    name='')
  final_fetches = [tfx.get_tensor(sess2.graph, name) for name in output_names]
  out_df = tfs.map_blocks(final_fetches, analyzed_df)

  for tnsr, output_colname in self.getOutputMapping():
    out_df = out_df.withColumnRenamed(tfx.op_name(sess.graph, tnsr), output_colname)
```
?

With IsolatedSession and GraphFunction, it's hard for anyone not intimately familiar with those classes to understand what's going on here. It makes it seem like you are just creating a GraphFunction object then getting back exactly what you put in. With more explicit code like above, it's easier to tell what exactly we're doing (pruning the graph into graphdef and re-importing). Also it seems shorter...

A few questions:
1/ I thought the tensorframes changes were done to remove the necessity for freezing+reimporting? But I may be wrong there (didn't really look at the tensorframes PR).
2/ In https://github.com/databricks/spark-deep-learning/blob/master/python/sparkdl/graph/builder.py#L91 (and potentially in the code I wrote above), do we need to check that no output has been stripped from the graph? Or is that not possible because the outputs were fed into the strip/freeze function and would be kept even if an output was "floating" and would have been stripped normally?

python/sparkdl/transformers/tf_tensor.py
```diff
@@ -0,0 +1,145 @@
...
+    keyword_only, HasInputMapping, HasOutputMapping, SparkDLTypeConverters,
+    HasTFGraph, HasTFHParams, HasTFCheckpointDir, HasExportDir, HasTagSet, HasSignatureDefKey)
+
+__all__ = ['TFTransformer']
+
+logger = logging.getLogger('sparkdl')
+
+class TFTransformer(Transformer, HasTFCheckpointDir, HasTFGraph,
+                    HasExportDir, HasTagSet, HasSignatureDefKey,
+                    HasTFHParams, HasInputMapping, HasOutputMapping):
```
anfeng: Where do you deal with batch size? Is that part of TFHParams?

python/tests/transformers/tf_tensor_test.py
```diff
@@ -0,0 +1,236 @@
...
+                     for row in df.select(output_col).toLocalIterator()])
+
+class TFTransformerTest(SparkDLTestCase):
+
+    def _get_rand_vec_df(self, num_rows, vec_size):
+        return self.session.createDataFrame(
+            Row(idx=idx, vec=np.random.randn(vec_size).tolist())
+            for idx in range(num_rows))
+
+    def test_checkpoint_reload(self):
```
anfeng: can we have a test program that construct transformer directly from a checkpoint file?

python/sparkdl/transformers/param.py
```diff
@@ -92,11 +99,45 @@ def getOutputCol(self):
...
+"""
+TensorFlow Specific Parameters
+"""
 class SparkDLTypeConverters(object):
 
     @staticmethod
+    def toTFGraph(value):
+        if isinstance(value, tf.Graph):
+            return value
+        elif isinstance(value, GraphFunction):
```
phi-dbq: Remove GraphFunction as it is internal API

thunterdb:
@phi-dbq I have merged #45, which should unblock you now. Let me know if you encounter some issues.

python/sparkdl/transformers/tf_tensor.py
```diff
@@ -0,0 +1,140 @@
...
+                self.output_renaming[tf_expected_colname] = output_colname
+
+            self.gfn = issn.asGraphFunction(feeds, fetches, strip_and_freeze=True)
+
+
+    def _transform(self, dataset):
+        self._convertInternal()
+
+        with IsolatedSession() as issn:
+            analyzed_df = tfs.analyze(dataset)
```
phi-dbq: With the latest tensorframes, we no longer need to rename input columns to match the graph tensor/operation names.

python/sparkdl/transformers/tf_tensor.py
```diff
@@ -0,0 +1,140 @@
...
+            _, fetches = issn.importGraphFunction(self.gfn, prefix='')
+            feed_dict = dict([(tnsr_name, col_name) for col_name, tnsr_name in self.getInputMapping()])
+            out_df = tfs.map_blocks(fetches, analyzed_df, feed_dict=feed_dict)
+
+            # We still have to rename output columns
+            for old_colname, new_colname in self.output_renaming.items():
```
phi-dbq: We still need to rename output columns. But that is consistent with the `map_row` API.

python/sparkdl/transformers/tf_tensor.py
```diff
@@ -0,0 +1,140 @@
...
+                assert _maybe_graph is not None
+
+            sig_def = None
+            if self.isDefined(self.signatureDefKey):
+                meta_graph_def = tf.train.export_meta_graph(issn.graph)
+                sig_def = tf.contrib.saved_model.get_signature_def_by_key(
+                    meta_graph_def, self.getSignatureDefKey())
+                assert sig_def is not None
+
+            feeds = []
```
phi-dbq: With the latest tensorframes, we don't have to collect input renaming mapping.

python/sparkdl/transformers/tf_tensor.py
```diff
@@ -0,0 +1,140 @@
...
+            self.output_renaming = {}
+            for tnsr, output_colname in self.getOutputMapping():
+                if sig_def:
+                    tnsr_name = tfx.tensor_name(issn.graph, tnsr)
+                    tnsr = sig_def.outputs[tnsr_name]
+                fetches.append(tfx.get_tensor(issn.graph, tnsr))
+                tf_expected_colname = tfx.op_name(issn.graph, tnsr)
+                self.output_renaming[tf_expected_colname] = output_colname
+
+            self.gfn = issn.asGraphFunction(feeds, fetches, strip_and_freeze=True)
```
phi-dbq: This a thin wrapper on top of `strip_and_freeze` utility function. 
I prefer to use it here as it hides us from the implementation details of said function, and gives the library maintainers an easy option to upgrade the underlying definition. 
It will be great if this library's exposure to TensorFlow API is limited.

python/sparkdl/transformers/tf_tensor.py
```diff
@@ -0,0 +1,140 @@
...
+        super(TFTransformer, self).__init__()
+        kwargs = self._input_kwargs
+        return self._set(**kwargs)
+
+
+    def _convertInternal(self):
+        assert self.isDefined(self.inputMapping) and self.isDefined(self.outputMapping), \
+            "inputMapping and outputMapping must be defined"
+
+        _maybe_graph = self.getTFGraph()
```
phi-dbq: The idea is that a user will provide one of the followings
1. a checkpoint directory
2. a tensorflow serving saved_model directory
3. a tensorflow graph object. 

We will convert all to an internal serializable format.

python/sparkdl/transformers/tf_tensor.py
```diff
@@ -0,0 +1,140 @@
...
+            sig_def = None
+            if self.isDefined(self.signatureDefKey):
+                meta_graph_def = tf.train.export_meta_graph(issn.graph)
+                sig_def = tf.contrib.saved_model.get_signature_def_by_key(
+                    meta_graph_def, self.getSignatureDefKey())
+                assert sig_def is not None
+
+            feeds = []
+            for _, tnsr_or_sig in self.getInputMapping():
+                if sig_def:
```
phi-dbq: When a `signature_def` is provided, the semantics of `inputMapping` and `outputMapping` change. 
Instead of mapping DataFrame columns to/from tensor/operation names, they now map columns to keys in the `signature_def` lookup table, which returns the actual tensor/operation names.

phi-dbq: We should probably make this process more transparent to users.

python/sparkdl/graph/utils.py
```diff
@@ -95,31 +95,49 @@ def get_tensor(graph, tfobj_or_name):
         'cannot locate tensor {} in current graph'.format(_tensor_name)
     return tnsr
 
-def as_tensor_name(name):
```
phi-dbq: There are multiple ways to get tensor/operation names in this utility module. 
We should definitely consolidate the functions `as_tensor_name` and `tensor_name`. 
Right now, the semantic of the two functions are:
1. `as_tensor_name`: return the tensor's name of the input (without necessarily check if it exists in any graph). 
2. `tensor_name`: return the tensor's name only if the input exists in the given graph.

python/sparkdl/transformers/param.py
```diff
@@ -92,11 +123,77 @@ def getOutputCol(self):
...
+            with IsolatedSession() as issn:
+                issn.importGraphFunction(value, prefix='')
+                g = issn.graph
+            return g
+        else:
+            raise TypeError("Could not convert %s to TensorFlow Graph" % type(value))
+
+    @staticmethod
+    def asColumnToTensorMap(value):
```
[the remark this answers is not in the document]
phi-dbq: I am using a converter function in `sparkdl.graph.utils` to convert the tensor/operation object into the corresponding operation name. If the operation is named `inputPlaceholder`, the tensor name is `inputPlaceholder:0`. TensorFrames expects the operation name.

python/sparkdl/transformers/param.py
```diff
@@ -92,11 +123,77 @@ def getOutputCol(self):
...
+                    continue
+                else:
+                    is_valid_type = False
+                    break
+            if is_valid_type:
+                return value
+        raise TypeError("Could not convert %s to TensorFlow Tensor" % type(value))
+
+    @staticmethod
+    def asTensorToColumnMap(value):
```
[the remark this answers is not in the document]
phi-dbq: I am using a converter function in `sparkdl.graph.utils` to convert the tensor/operation object into the corresponding operation name. If the operation is named `inputPlaceholder`, the tensor name is `inputPlaceholder:0`. TensorFrames expects the operation name.

python/tests/transformers/tf_tensor_test.py
```diff
@@ -0,0 +1,301 @@
...
+            # Model definition ends
+
+            sess.run(w.initializer)
+
+            sig_inputs = {
+                'input_sig': tf.saved_model.utils.build_tensor_info(x)}
+            sig_outputs = {
+                'output_sig': tf.saved_model.utils.build_tensor_info(z)}
+
+            serving_sigdef = tf.saved_model.signature_def_utils.build_signature_def(
```
phi-dbq: The shape and data types of the model is stored as well.
```
sigdef: {u'prediction_signature': inputs {
  key: "input_sig"
  value {
    name: "tnsrIn:0"
    dtype: DT_DOUBLE
    tensor_shape {
      dim {
        size: -1
      }
      dim {
        size: 17
      }
    }
  }
}
outputs {
  key: "output_sig"
  value {
    name: "tnsrOut:0"
    dtype: DT_DOUBLE
    tensor_shape {
      dim {
        size: -1
      }
    }
  }
}
}
```

python/sparkdl/transformers/tf_tensor.py
sueann: I strongly disagree. Calling strip_and_freeze directly also hides the details of the function, so why do we need another wrapper on top of fit? It's very time-consuming to understand the IsolatedSession code which means ample room for mistakes in coding and debugging this code here.

python/sparkdl/transformers/utils.py
```diff
@@ -36,3 +34,147 @@ class ImageNetConstants:
...
     INPUT_SHAPE = (299, 299)
     NUM_OUTPUT_FEATURES = 131072
+
+
+class TFInputGraph(object):
+    """
+    An opaque serializable object containing TensorFlow graph.
+    """
+    # TODO: for (de-)serialization, the class should correspond to a ProtocolBuffer definition.
+    def __init__(self, graph_def, input_mapping, output_mapping):
```
sueann: Talked offline -- TFInputGraph should be a super thin wrapper containing graph_def. It's constructor really shouldn't be called by the user and they should use TFInputGraphBuilder below -- we should at least document this and warn that the API could change for the constructor here.

python/sparkdl/transformers/param.py
```diff
@@ -92,11 +97,58 @@ def getOutputCol(self):
         return self.getOrDefault(self.outputCol)
 
 
-# New in sparkdl
-
+"""
+TensorFlow Specific Parameters
```
sueann: Remove this comment. As the name suggests, the converters are mean to be for SparkDL generally (though they are mostly for TensorFlow, there is one that's generically model name at the bottom, for example).

python/tests/transformers/tf_tensor_test.py
```diff
@@ -0,0 +1,323 @@
...
+                'output_sig': output_col
+            })
+
+        # Build the transformer from exported serving model
+        # We are not using signatures, thus must provide tensor/operation names
+        trans_no_sig = TFTransformer(
+            tfInputGraph=TFInputGraphBuilder.fromSavedModel(
+                saved_model_dir, tag_set=serving_tag, signature_def_key=None),
+            inputMapping={
+                input_col: 'tnsrIn'
```
sueann: from our earlier discussion, i thought this would look like 
```
  inputMapping={
    input_col: 'input_sig',
  }
```
no? If `input_col: 'tnsrIn'` is what's expected, we don't need the helper functions or the sigDefKey logic to get the inputMapping and outputMapping, do we?

python/tests/transformers/tf_tensor_test.py
```diff
@@ -0,0 +1,323 @@
...
+        saved_model_root = tempfile.mkdtemp()
+        saved_model_dir = os.path.join(saved_model_root, 'saved_model')
+        serving_tag = "serving_tag"
+        serving_sigdef_key = 'prediction_signature'
+
+        builder = tf.saved_model.builder.SavedModelBuilder(saved_model_dir)
+        with tf.Session(graph=tf.Graph()) as sess:
+            # Model definition: begin
+            x = tf.placeholder(tf.float64, shape=[None, vec_size], name='tnsrIn')
+            #x = tf.placeholder(tf.float64, shape=[None, vec_size], name=input_col)
```
sueann: remove if not used.

python/tests/transformers/tf_tensor_test.py
```diff
@@ -0,0 +1,323 @@
...
+
+        # Setup saved model export directory
+        saved_model_root = tempfile.mkdtemp()
+        saved_model_dir = os.path.join(saved_model_root, 'saved_model')
+        serving_tag = "serving_tag"
+        serving_sigdef_key = 'prediction_signature'
+
+        builder = tf.saved_model.builder.SavedModelBuilder(saved_model_dir)
+        with tf.Session(graph=tf.Graph()) as sess:
+            # Model definition: begin
```
sueann: let's have all the `test_build_...` tests in this class use the same graph. can we refactor so we have a helper function constructing the graph? (i know it's annoying with session etc, but it's possible.)

python/tests/transformers/tf_tensor_test.py
```diff
@@ -0,0 +1,323 @@
...
+                q: 'outcol_q'
+            })
+        final_df = transfomer.transform(analyzed_df)
+        p_out_tgt = grab_df_arr(final_df, 'outcol_p')
+        q_out_tgt = grab_df_arr(final_df, 'outcol_q')
+
+        self.assertTrue(np.allclose(p_out_ref, p_out_tgt))
+        self.assertTrue(np.allclose(q_out_ref, q_out_tgt))
+
+    def test_mixed_keras_graph(self):
```
sueann: what are we trying to accomplish with this test? is the goal to test a larger graph (doesn't look that big?)? or test compatibility with keras?

python/sparkdl/transformers/param.py
```diff
@@ -92,11 +97,58 @@ def getOutputCol(self):
...
+            return TFInputGraphBuilder.fromGraph(value)
+        elif isinstance(value, tf.GraphDef):
+            return TFInputGraphBuilder.fromGraphDef(value)
+        else:
+            raise TypeError("Could not convert %s to TFInputGraph" % type(value))
+
+    @staticmethod
+    def asColumnToTensorMap(value):
+        if isinstance(value, dict):
+            strs_pair_seq = [(k, tfx.as_op_name(v)) for k, v in value.items()]
```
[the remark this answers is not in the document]
phi-dbq: I think apart from `feetds` and `fetches` in `session.run`, all the references use `tf.Operation` names. I will update the docs.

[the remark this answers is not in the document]
sueann: Is it possible though to have two tensors with the same op name in the mappings though?

python/tests/transformers/tf_tensor_test.py
phi-dbq: The motivation is to see if we can build TFTransformer from TF + Keras mixed layers.

sueann: It's good to know that works, though I am not sure why we would question that it works in this context (i.e. this doesn't seem like the right place to test that vs tensorframes and/or the strip_and_freeze unit tests - yes i know we don't have any). in any case, i'd leave it in for now but add a TODO comment to remove it once we add the keras transformer.

python/sparkdl/transformers/param.py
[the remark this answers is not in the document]
phi-dbq: It might happen for ops that return multiple tensors. These operations are not common for inputs and outputs in TensorFlow graphs. I will update the documentation to clarify.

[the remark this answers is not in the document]
sueann: Then let's make them tensor names instead of op names (we do that in other transformers in sparkdl anyway). Any reason why op names are much more preferable?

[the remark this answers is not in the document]
phi-dbq: We might actually not be able to do this since `TensorFrames` name the output column to the operation name ([we do this in tf_image module](https://github.com/databricks/spark-deep-learning/blob/master/python/sparkdl/transformers/tf_image.py#L140-L144)). 
If there are two tensor outputs from the same operation, there will be a column name conflict.

python/sparkdl/transformers/param.py
```diff
@@ -107,18 +159,74 @@ def toStringOrTFTensor(value):
...
+    def getOutputMapping(self):
+        return self.getOrDefault(self.outputMapping)
+
+
+class HasInputMapping(Params):
+    """
+    Mixin for param inputMapping: ordered list of ('inputColName', 'inputTensorName') pairs
+    """
+    inputMapping = Param(Params._dummy(), "inputMapping",
+                         "Mapping input DataFrame column names to :class:`tf.Tensor` objects",
```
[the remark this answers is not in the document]
phi-dbq: I will document the actual converted type here and indicate what types are automatically converted.

python/tests/transformers/tf_tensor_test.py
phi-dbq: That sounds very useful. Shall we make a separate PR to refactor the `test_build_...` tests?

python/sparkdl/transformers/param.py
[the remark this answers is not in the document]
sueann: My comment was, the user should give a mapping {dataframe column name : tensor name} not {data frame column name : tensor object} but the documentation says otherwise. They don't need to know about internal conversions.

python/sparkdl/transformers/param.py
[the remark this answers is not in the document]
sueann: I see. We only output one tensor/op in the image transformers so this is not a problem there. 1/ Could we put a check in the output mappings and warn if there are two tensors with the same op name? 
2/ Let's make sure users only deal with tensor names and op names are internal (not sure if this is already true in the code).

python/tests/transformers/tf_tensor_test.py
sueann: I'd just refactor here. It's not a adding another feature, just making the code more readable and debuggable as it should have been.

python/sparkdl/transformers/tf_tensor.py
```diff
@@ -0,0 +1,92 @@
...
+        self.setParams(**kwargs)
+
+    @keyword_only
+    def setParams(self, tfInputGraph=None, inputMapping=None, outputMapping=None, tfHParms=None):
+        """
+        setParams(self, tfInputGraph=None, inputMapping=None, outputMapping=None, tfHParms=None)
+        """
+        super(TFTransformer, self).__init__()
+        kwargs = self._input_kwargs
+        _maybe_gin = SparkDLTypeConverters.toTFInputGraph(tfInputGraph)
```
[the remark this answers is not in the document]
phi-dbq: Refactored so that the conversion logic lives here.

python/tests/transformers/tf_tensor_test.py
phi-dbq: When the user specifically indicated not to use SignatureDef by setting `signature_def_key=None`, we expect the operation names to be provided.

python/sparkdl/transformers/param.py
```diff
@@ -50,10 +56,8 @@ class HasInputCol(Params):
     Mixin for param inputCol: input column name.
     """
 
-    inputCol = Param(Params._dummy(), "inputCol", "input column name.", typeConverter=TypeConverters.toString)
-
-    def __init__(self):
-        super(HasInputCol, self).__init__()
+    inputCol = Param(Params._dummy(), "inputCol", "input column name.",
```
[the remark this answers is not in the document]
phi-dbq: They are YAPF'ed along with the rest of this module. Looks like PySpark could improve on format/style. Will convert them back to PySpark.

python/sparkdl/transformers/tf_tensor.py
```diff
@@ -0,0 +1,111 @@
...
+class TFTransformer(Transformer, HasTFInputGraph, HasTFHParams, HasInputMapping, HasOutputMapping):
+    """
+    Applies the TensorFlow graph to the array column in DataFrame.
+
+    Restrictions of the current API:
+
+    We assume that
+    - All graphs have a "minibatch" dimension (i.e. an unknown leading
+      dimension) in the tensor shapes.
+    - Input DataFrame has an array column where all elements have the same length
```
sueann: all rows (not elements)?

python/sparkdl/transformers/tf_tensor.py
```diff
@@ -0,0 +1,111 @@
...
+    @keyword_only
+    def setParams(self, tfInputGraph=None, inputMapping=None, outputMapping=None, tfHParms=None):
+        """
+        setParams(self, tfInputGraph=None, inputMapping=None, outputMapping=None, tfHParms=None)
+        """
+        super(TFTransformer, self).__init__()
+        kwargs = self._input_kwargs
+        # The set of parameters either come from some helper functions,
+        # in which case type(_maybe_gin) is already TFInputGraph.
+        _maybe_gin = tfInputGraph
```
sueann: this variable is unnecessary

python/sparkdl/transformers/tf_tensor.py
```diff
@@ -0,0 +1,111 @@
...
+        if isinstance(_maybe_gin, TFInputGraph):
+            return self._set(**kwargs)
+
+        # Otherwise, `_maybe_gin` needs to be converted to TFInputGraph
+        # We put all the conversion logic here rather than in SparkDLTypeConverters
+        if isinstance(_maybe_gin, TFInputGraphBuilder):
```
sueann: this shouldn't be possible anymore

sueann:
@phi-dbq, left you a few more comments skimming the latest but I'm not sure if I'm supposed to review again yet. Let me know when it's ready.

phi-dbq:
Closing this PR in favor of the stacked PR holistic view
https://github.com/phi-dbq/spark-deep-learning/pull/12

PR #42, opened 2017-08-10
Linter

phi-dbq:

Add prospector for style checking, etc.


thunterdb:
@phi-dbq thanks! Can you please document in the README (for example in running teh unit tests) that you need prospector installed with a link to the page of that tool?
Also, I am surprised it has not found anything... Are there some good rules we have not turned on?

PR #46, opened 2017-08-26
Add support for Reading GIF files

seanpquig:
We have been using this library over at [GIPHY|https://github.com/Giphy] and love it.  We had to adapt it to work with GIFs, and we thought we'd share some of the changes with the community.  It includes:

- [x] `readGifs` function that reads a directory of GIFs and splits them out into individual frames/images that can be fed into `InceptionV3` and other models.
- [x] supporting unit tests mimicking tests in `TestReadImages`
- [x] updates to `.gitignore`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.log
 *.pyc
 build/*.jar
+.coverage
 
 docs/_site
 docs/api
@@ -17,6 +18,9 @@ lib_managed/
 src_managed/
 project/boot/
 project/plugins/project/
+
+# spark
+metastore_db
 
 # intellij
 .idea/
--- a/python/sparkdl/image/imageIO.py
+++ b/python/sparkdl/image/imageIO.py
@@ -34,6 +34,10 @@ imageSchema = StructType([StructField("mode", StringType(), False),
                           StructField("nChannels", IntegerType(), False),
                           StructField("data", BinaryType(), False)])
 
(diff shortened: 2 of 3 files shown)


thunterdb:
Hello @seanpquig thank you very much for this contribution, we will be happy to add support for GIFs to Deep Learning Pipelines. 

I have some design questions about the new schema added for GIF which we should be able to resolve without too much change on your side.

To give some context, we are in the process of consolidating different image processing solutions around the image schema described in `python/sparkdl/image/imageIO.py` and I believe that we can add some extra fields to the image schema to handle gifs, without having to create a separate schema. Before offering some changes, I would like to understand a bit more some of the use case: from looking at `gifSchema`, it looks like you do not use the fact that frames in a GIF are ordered and you simply store them independently in a dataframe? Do you foresee such a use case of keeping all the frames together?

seanpquig:
Hey @thunterdb.  I tried to take a minimal and flexible approach and have the gif schema be per frame and identical to the image schema with an additional `frameNum` field to keep track of ordering.  This has allowed us to write our own custom functions and processing to do things like frame sampling, averaging model predictions across frames, and investigating ordering effects.

I think combining in a common schema could be great as long as it doesn't sacrifice any information and the ability to access individual frames.  Perhaps images could be loosely modeled as special case of a GIF that has a single frame.  Your thoughts?

Issue #53, opened 2017-09-29
[PR:#45] To Avoid collect tranning data to driver and broadcast  them 


allwefantasy:
I check the last PR in spark-deep-learning is [KerasImageFileEstimator(https://github.com/databricks/spark-deep-learning/pull/35), and when i review the code, i find it will collect all trainning data to driver and then broadcast to executors. This means all tranning data should fit in one server memory and it will definitely not work in real world especially when deep learning is a  data-hungry ML algrithom. 

Maybe we can write  tranning data to  a distributed message queue eg. Kafka, then we invoke tf queue to recevie data from kafka and consume data from tf queue when tf session starts.

```python
class KerasImageFileEstimator(Estimator, HasInputCol, HasInputImageNodeName,
                              HasOutputCol, HasOutputNodeName, HasLabelCol,
                              HasKerasModel, HasKerasOptimizer, HasKerasLoss,
                              CanLoadImage, HasOutputMode，DistributedModel="ParamsParallel", KafkaServer="127.0.0.1"):
```

We also can put data in HDFS as optional , but Message Queue sees to be a perfect choice.

PR #56, opened 2017-10-13
NLP support

allwefantasy:
What changes are proposed in this pull request?

Creating a Spark MLlib Estimator API which can integrated with tensorflow code, with a reference implementation.
Creating a Spark MLlib Transformer convert text column to 2-D vector which can be feeded to CNN/LSTM directly.

It provides a taste of how to process text  in a DataFrame and use them to train a NLP  model developed by tensorflow.
Also fix issue: https://github.com/databricks/spark-deep-learning/issues/53

The changes consist of these components.

TFTextFileEstimator/TFTextTransformer
New shard params 

How is this patch tested?

*  Unit tests
* Manual tests


phi-dbq:
Thank you for contributing to the project @allwefantasy!

allwefantasy:
The CI fails because there is no kafka lib in the env. Is there something i can do to fix this ?

thunterdb:
@allwefantasy thank you very much for the contribution. I will have more comments for the estimator, so would you mind splitting your PR into the transformer part and into the estimator?

Also, I see that the transformer is embedding Word2Vec. Have you considered chaining them in a pipeline instead? 
https://spark.apache.org/docs/2.1.1/ml-pipeline.html

Regarding kafka, you should be able to add it in this file:
https://github.com/databricks/spark-deep-learning/blob/master/python/requirements.txt

allwefantasy:
@thunterdb  TFTextTransformer is a tool like StringIndexer in MLlib, which we can use to transform the dataframe and feed the new dataframe to TFTextFileEstimator. It seems no need to split into two PRs.

Using word2vec is in order to compute a Map which contains the mapping of word to vector. We do not need the word2vec model's transform function.

PR #62, opened 2017-10-18
Scala API for image featurizer

MrBago:
We're anticipating switching the image schema to `spark.image.ImageSchema` (https://issues.apache.org/jira/browse/SPARK-21866).

Until that schema is merged we'll be depending on `microsoft/spark-image`. There is a small difference in the schema that'll need to be fixed when the change happens.


phi-dbq:
@MrBago, thank you for contributing the Scala API and reference implementation of DeepImageFeaturizer! 
Shall we break this holistic view PR into one for imageUtils and one for the featurizer implementation? 
Given that we also introduced an external dependency on https://github.com/Microsoft/spark-images, shall we add a few tests regarding our assumptions on the functionality in there that we rely on?

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+        w += 1
+      }
+      h += 1
+    }
+
+    val scaledImg = srcImg.getScaledInstance(tgtHeight, tgtWidth, Image.SCALE_AREA_AVERAGING)
+    val tgtImg = new BufferedImage(tgtHeight, tgtWidth, BufferedImage.TYPE_3BYTE_BGR)
+
+    // Draw the image on to the buffered image
+    val graphic = tgtImg.createGraphics()
```
phi-dbq: As we discussed offline, one concern of having this function is that it might depend on the underlying GUI rendering system. If such system is deliberately removed, it could fail. Could you please check at least if the Linux implementation has such dependencies?

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+    graphic.drawImage(scaledImg, 0, 0, null)
```
phi-dbq: Would the drawImage have any primary or secondary interaction with the framebuffer?

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+   *   val myResizeUdf = udf(resizeSPImage(height, width, 3) _, ImageSchema.columnSchema)
+   *
+   * @param tgtHeight Desired height of the image.
+   * @param tgtWidth Desired width of the image.
+   * @param tgtChannels Target number of channels (must be 3 for now).
+   * @param r Image to resize.
+   * @return Row, The resized image.
+   */
+  def resizeSPImage(tgtHeight: Int, tgtWidth: Int, tgtChannels: Int)(r: Row): Row = {
+    val height = r.getInt(1)
```
phi-dbq: As we discussed this offline, can we add documentation for the implicit dependency on `spark.ml.image` ImageSchema?

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+      tgtChannels: Int,
+      imageData: Array[Byte]) = {
+    require(tgtChannels == 3, "output must have 3 channels.")
+    val srcImg = new BufferedImage(srcHeight, srcWidth, BufferedImage.TYPE_3BYTE_BGR)
+
+    var offset, h = 0
+    while (h < srcHeight) {
+      var w = 0
+      while (w < srcWidth) {
+        val (r, g, b): (Byte, Byte, Byte) = srcChannels match {
```
phi-dbq: Let's refactor this code into two loops, one for `srcChannels == 1` and one for `srcChannels == 3`. 
This way we only have to deal with the condition once outside the loop, making the code more performant.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+
+        val color = new Color(r & 0xff, g & 0xff, b & 0xff)
+        srcImg.setRGB(h, w, color.getRGB)
+        offset += srcChannels
+        w += 1
+      }
+      h += 1
+    }
+
+    val scaledImg = srcImg.getScaledInstance(tgtHeight, tgtWidth, Image.SCALE_AREA_AVERAGING)
```
phi-dbq: The contract, as I believe, is that `DeepImageFeaturizer` should behave the same whether we call it from Scala or from Python. Let's find out a way to make sure they do and add tests.
Thank you!

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+    val tgtImg = new BufferedImage(tgtHeight, tgtWidth, BufferedImage.TYPE_3BYTE_BGR)
+
+    // Draw the image on to the buffered image
+    val graphic = tgtImg.createGraphics()
+    graphic.drawImage(scaledImg, 0, 0, null)
+    graphic.dispose()
+
+    val decoded = new Array[Byte](tgtHeight * tgtWidth * tgtChannels)
```
phi-dbq: There are quite a few round trips for the image bytes. 
If there is an existing library that does this efficiently (which I believe there is), I think it is worth it to incorporate it as a dependency. 

In addition, since we already depend on TensorFlow, we may consider reusing some of its image utilities.

MrBago:
I split this PR into multiple PRs, starting with #64.

phi-dbq:
Okay, got it. I will mark my comments for the image-utils to #64. 
Thank you!

PR #64, opened 2017-10-18
ImageUtils for use in DeepImageFeaturizer

MrBago:
Added utility to resize images for use with Image Transforms that expect images of a given size.


MrBago:
Though this PR reports at > 500 lines, most of that is the test image `biggerImage.raw`.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+   * @return Array[Byte] of image data, if the input was BGR or 1 channel, the output will be BGR.
+   */
+  private[sparkdl] def resizeImage(
+      srcHeight: Int,
+      srcWidth: Int,
+      srcChannels: Int,
+      tgtHeight: Int,
+      tgtWidth: Int,
+      tgtChannels: Int,
+      imageData: Array[Byte]) = {
```
thunterdb: return type

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+    require(tgtChannels == 3, "output must have 3 channels.")
```
thunterdb: also put `tgtChannels` in the text of the assertion

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+    val srcImg = new BufferedImage(srcHeight, srcWidth, BufferedImage.TYPE_3BYTE_BGR)
+
+    var offset, h = 0
+    while (h < srcHeight) {
```
thunterdb: can you separate the code that creates `BufferedImage` in a function? I am sure there will be other usages.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+      var w = 0
+      while (w < srcWidth) {
+        val (r, g, b): (Byte, Byte, Byte) = srcChannels match {
```
thunterdb: this is going to do pattern matching and allocation of an object. You can define 3 `var`s instead.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+            val g = imageData(offset + 1)
+            val r = imageData(offset + 2)
+            (r, g, b)
+          case _ =>
+            require(false, "`srcChannels` must be 1 or 3.")
+            // make type check happy
+            (0, 0, 0)
+        }
+
+        val color = new Color(r & 0xff, g & 0xff, b & 0xff)
```
thunterdb: I wonder if you can create the color outside and just set the channels with getters.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+
+    val scaledImg = srcImg.getScaledInstance(tgtHeight, tgtWidth, Image.SCALE_AREA_AVERAGING)
+    val tgtImg = new BufferedImage(tgtHeight, tgtWidth, BufferedImage.TYPE_3BYTE_BGR)
+
+    // Draw the image on to the buffered image
+    val graphic = tgtImg.createGraphics()
+    graphic.drawImage(scaledImg, 0, 0, null)
+    graphic.dispose()
+
+    val decoded = new Array[Byte](tgtHeight * tgtWidth * tgtChannels)
```
thunterdb: same thing, can you separate this code in a function?

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+    offset = 0
+    h = 0
+    while (h < tgtHeight) {
+      var w = 0
+      while (w < tgtWidth) {
+        val color = new Color(tgtImg.getRGB(h, w))
```
thunterdb: if you can, I wonder if there is a way not to allocate objects in the loop.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+  def resizeSPImage(tgtHeight: Int, tgtWidth: Int, tgtChannels: Int)(r: Row): Row = {
+    val height = r.getInt(1)
+    val width = r.getInt(2)
+    val nChannels = r.getInt(3)
+    if ((nChannels == tgtChannels) && (height == tgtHeight) && (width == tgtWidth)) {
+      r
+    } else {
+      val data = r.getAs[Array[Byte]](5)
+      val resized = resizeImage(height, width, nChannels, tgtHeight, tgtWidth, tgtChannels, data)
+      // TODO: udpate mode to be Int when spark.image is merged.
```
thunterdb: `spark.ml.image`

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+      val mode = "CV_8UC3"
+      Row(null, tgtHeight, tgtWidth, tgtChannels, mode, resized)
```
thunterdb: `null`? I wonder if we should carry it over instead. This is not specified in any case.

src/test/resources/smallerImage.raw
```diff
@@ -0,0 +1 @@
...
,&>7%50
@<(=8+1*)(
$&
45+CDDWYZ���vwy���~�}������������bec���DDE! 
;=J���������qtq��~������non���>3";5#5.
3,93!5094$ID4:8*KJ@GFA?=0@>5;;6bc_WXXrur���������|}������������ZZQOQJ./)#"!15AT[vbl}VYVsvn������eicy}x������H8!A9(71!<4$74'1/!DB4KJACE:ikhOOMedc[[Zijjy{zhjh���UWS�����������ehe������NL@;5'&"""!25CLRg���������VYQ������x|}prp���N8
51#>8(G@1;;264*ZVPTUSYZTbccZ[XssrUUQ���hjgac]���}��������rwr���xywED:GC6OJ77/&"('
*+-38D{��lni���������y{x��������\<D<.A>.II@A@=JJDeegKKFcd`RRIiigz}�^\Uuus\`^fgc���Z^__bY���klhLLAWWMjj_ED@52)7/:050
-/117@������moj}}w���������eha���lB
J:)YVGbc_PPJdd`HHDNOFhikKI>{zt���\ZWgif���KNGvxo���bb^<<1omcjjdMMEFF>QPMmkfOH9B;)<3!2. ikppsr��y������tvuoqi���������uHN8!>9-UUNKLFZ[X>>6pnf���qrqIG>jia��~MOK\^[�HIFEC9ie`qpk_]YKNDijf���llkttsMKAgaUb_Tcc]|{v������ihdfga������w{~WXTyzv{P"ZD(BA8KH:_^XjhgXZXGE;VUI����CF=UVI{zztvw9</_`WffbhihRSItuo������ehiZZS������WWRomb������jpk[^R������OSKpsn����������W'hDPIAVTMMKGJJ@\^Y���GHFJF?yww���NNJB?5kgbdba<<1BB5loe���XZWXXU������wzz]_[������x{xWWQwws���lnn^b[z}u������dea�������^+zQ#P?'``XijjVXX53'POEtsodgjEGBRVLomhdddRRIfe]nngkjhgigQSHfjd������_b\������efffhdvxu���sy{`a[������a`\������bd\���tvs
```
thunterdb: I cannot see the image, but are you sure they have a good license?

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
phi-dbq: +1
Also, can we move srcChannels outside the while loop? If is determined per image, and matching them inside the loop incurs some overhead.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+          case 1 =>
+            val i = imageData(offset)
+            (i, i, i)
+          case 3 =>
+            val b = imageData(offset)
+            val g = imageData(offset + 1)
+            val r = imageData(offset + 2)
+            (r, g, b)
+          case _ =>
+            require(false, "`srcChannels` must be 1 or 3.")
```
phi-dbq: Same as before, this test is better done outside the while-loop. 
And we should report the value of the faulty `srcChannels` parameter.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
MrBago: I looked into it after we talked last week, and Color is immutable. I don't know why Color, or all things, is immutable in java...

I can replace `rgb = Color(red, green, blue).getRGB` with `rgb = (red << 16 | green << 8 | blue)`. But I think that's kinda gross.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,131 @@
...
+        w += 1
+      }
+      h += 1
+    }
+
+    val scaledImg = srcImg.getScaledInstance(tgtHeight, tgtWidth, Image.SCALE_AREA_AVERAGING)
+    val tgtImg = new BufferedImage(tgtHeight, tgtWidth, BufferedImage.TYPE_3BYTE_BGR)
+
+    // Draw the image on to the buffered image
+    val graphic = tgtImg.createGraphics()
```
phi-dbq: As we discussed offline, this might potentially have dependencies on the platform's GUI implementation, which might cause trouble if they are disabled in some deployments.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
MrBago: done

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
MrBago: done

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
MrBago: done

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
MrBago: done

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
MrBago: Color objects can't be re-used, the only way we can avoid create color objects is the manipulate the bits directly, ie `rgb = (red << 16 | green << 8 | blue)`.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
MrBago: I don't think carrying over is the right thing to do, once we've manipulated it, this image is different than the one at `origin`. But, admittedly this really depends on what the purpose/definition of the origin field is :).

src/test/resources/smallerImage.raw
MrBago: I've droped this raw files altogether, and switched to reusing images we already have in the repo.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
MrBago: done

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,142 @@
...
+          | image of size ($height, $width, $channels).
+       """.stripMargin
+    )
+    val image = new BufferedImage(width, height, BufferedImage.TYPE_3BYTE_BGR)
+
+    var offset, h = 0
+    var r, g, b: Byte = 0
+    while (h < height) {
+      var w = 0
+      while (w < width) {
```
phi-dbq: (Migrated from the comments on https://github.com/databricks/spark-deep-learning/pull/62)
Shall we refactor this code into two loops, one for `channels == 1` and one for `channels == 3`. 
This way we only have to deal with the condition once outside the loop, making the code more performant.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,142 @@
...
+
+    if ((nChannels == tgtChannels) && (height == tgtHeight) && (width == tgtWidth)) {
+      spImage
+    } else {
+      val srcImg = spImageToBufferedImage(spImage)
+      val tgtImg = new BufferedImage(tgtWidth, tgtHeight, BufferedImage.TYPE_3BYTE_BGR)
+
+      // Draw srcImage onto resized (usually smaller) tgtImg
+      val graphic = tgtImg.createGraphics()
+      val scaledImg = srcImg.getScaledInstance(tgtWidth, tgtHeight, Image.SCALE_AREA_AVERAGING)
```
phi-dbq: The contract, as I believe, is that `DeepImageFeaturizer` should behave the same whether we call it from Scala or from Python. Let's find out a way to make sure they do and add tests.
Thank you!

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,142 @@
...
+    val nChannels = ImageSchema.getNChannels(spImage)
+
+    if ((nChannels == tgtChannels) && (height == tgtHeight) && (width == tgtWidth)) {
+      spImage
+    } else {
+      val srcImg = spImageToBufferedImage(spImage)
+      val tgtImg = new BufferedImage(tgtWidth, tgtHeight, BufferedImage.TYPE_3BYTE_BGR)
+
+      // Draw srcImage onto resized (usually smaller) tgtImg
+      val graphic = tgtImg.createGraphics()
```
phi-dbq: As we discussed offline, one concern of having this function is that it might depend on the underlying GUI rendering system. If such system is deliberately removed, it could fail. Could you please check at least if the Linux implementation has such dependencies?

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,142 @@
...
+      val scaledImg = srcImg.getScaledInstance(tgtWidth, tgtHeight, Image.SCALE_AREA_AVERAGING)
+      graphic.drawImage(scaledImg, 0, 0, null)
```
phi-dbq: Would the drawImage have any primary or secondary interaction with the framebuffer?

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,142 @@
...
+   *
+   * @param image Java BufferedImage.
+   * @return Row image in spark.ml.image format with 3 channels in BGR order.
+   */
+  private[sparkdl] def spImageFromBufferedImage(image: BufferedImage): Row = {
+    val channels = 3
+    val height = image.getHeight
+    val width = image.getWidth
+
+    val decoded = new Array[Byte](height * width * channels)
```
phi-dbq: There are quite a few round trips for the image bytes. 
If there is an existing library that does this efficiently (which I believe there is), I think it is worth it to incorporate it as a dependency. 

In addition, since we already depend on TensorFlow, we may consider reusing some of its image utilities.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
MrBago: I think the jvm should handle the loop unswitching here. My understanding is that loop unswitching only has marginal benefits even in non-compiled languages, http://labs.septeni-technology.jp/none/loop-unswitching-a-performance-appraisal/.

My suggestion is that we focus on readability in this PR and push performance improvements to a separate PR which can include benchmarks.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
MrBago: I think in the long term we'll make sure that the resizing is the same by having the python code call out to the scala implementation. Because of that I think we should focus on correctness here. I'm not sure what scaling method we're using in the python code, but we should use the AREA scaling method in both python and in scala, http://tanbakuchi.com/posts/comparison-of-openv-interpolation-algorithms/#Downsampling-comparison.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
MrBago: We don't use any of the classes/objects that require GUI rendering, http://www.oracle.com/technetwork/articles/javase/headless-136834.html.

To make sure we can run all the image resizing stuff in headless mode, I've added "headless=ture" to JvmOptions, https://github.com/databricks/spark-deep-learning/pull/64/files#diff-fdc3abdfd754eeb24090dbd90aeec2ceR76 and I test the awt.headless property at test time.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
MrBago: I looked into using tensorflow, but my understanding is that tf.image is not yet available in the java api. I'm open to using an outside library to do some of this stuff, but that seems like a performance optimization we can add later once we have a better sense of where the bottlenecks are.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,142 @@
...
+        decoded(offset) = color.getBlue.toByte
+        decoded(offset + 1) = color.getGreen.toByte
+        decoded(offset + 2) = color.getRed.toByte
+        offset += channels
+        w += 1
+      }
+      h += 1
+    }
+    // TODO: udpate mode to be Int when spark.ml.image is merged.
+    Row(null, height, width, channels, "CV_U8C3", decoded)
```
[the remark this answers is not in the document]
MrBago: I'm not sure, what did you have in mind? Something like:
```
def rowImage(origin, height, width, channels, mode, data) = Row(origin, height, width, channels, mode, data)
```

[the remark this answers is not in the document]
smurching: Yeah, it looks a bit redundant but I think it makes it clearer what the row fields are supposed to represent & would simplify updating the code if there's a change to `ImageSchema`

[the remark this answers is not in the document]
smurching: I'm ok with not changing the existing code either though, this seems like a small thing.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
phi-dbq: Shall we check out TensorFlow's Java API (with this [image labeling example](https://github.com/tensorflow/tensorflow/blob/r1.3/tensorflow/java/src/main/java/org/tensorflow/examples/LabelImage.java)).
I believe the standard operations are available in Java, specifically,

- Image ops are [implemented in C++](https://github.com/tensorflow/tensorflow/blob/v1.3.1/tensorflow/core/ops/image_ops.cc) and made [available in Python](https://github.com/tensorflow/tensorflow/blob/v1.3.1/tensorflow/python/ops/image_ops_impl.py).
- Tensor [slicing and transformation ops](https://www.tensorflow.org/api_guides/python/array_ops) should be useful, too.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
phi-dbq: I agree that loop optimizations (e.g. unswitching), might not always provide the best performance. 
However, in this case, the [JVM loop unswitching bug](https://github.com/rholder/jvm-loop-unswitching-bug) might have a chilling effect on JVM to apply loop unswitching automatically.

phi-dbq: The argument in the [cited article](http://labs.septeni-technology.jp/none/loop-unswitching-a-performance-appraisal/) is flawed as the micro-benchmark is not correctly conducted.

phi-dbq: I agree that readability is important. In this case though, splitting it into two loops does not hinder readability as we don't have to deal with BGR conversion for mono channel images. 
I will leave the readability discussion to @sueann or @thunterdb.

phi-dbq: I will be fantastic if we can get another PR to focus on performance benchmarking.
Data ingestion is on the critical path for massively distributed deep learning workload.

sueann: We can worry about performance once we have some users... :-)

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
sueann: Would this be solved by having the python version of `DeepImageFeaturizer` call into the scala version? If that's simple, we can do it as part of the `DeepImageFeaturizer` persistence work. There is no persistence support right now so we don't need to worry about it. Note that `DeepImagePredictor` will continue to use the python implementation due to the dependency on keras.

The python version of resizing uses `imageIO.resizeImage` which uses PIL.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,144 @@
...
+        w += 1
+      }
+      h += 1
+    }
+    image
+  }
+
+
+  /**
+   * Takes a Java BufferedImage and returns a Row Image (spImage). The
```
sueann: Nit: The?

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,144 @@
...
+          case 3 =>
+            b = imageData(offset)
+            g = imageData(offset + 1)
+            r = imageData(offset + 2)
+          case _ =>
+            require(false, s"`Channels` must be 1 or 3, got $channels.")
+        }
+
+        val color = new Color(r & 0xff, g & 0xff, b & 0xff)
+        image.setRGB(w, h, color.getRGB)
```
sueann: would using the vector version be faster?

sueann: can leave a TODO to try if it's too much work

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,144 @@
...
+    val image = new BufferedImage(width, height, BufferedImage.TYPE_3BYTE_BGR)
+
+    var offset, h = 0
+    var r, g, b: Byte = 0
+    while (h < height) {
+      var w = 0
+      while (w < width) {
+        channels match {
+          case 1 =>
+            b = imageData(offset)
```
sueann: is this how one would normally convert a 1-channel image to a 3-channel (all same values)? it's kinda nice that you can feed in 1-channel images to DeepImageFeaturizer as a consequence of resizing :P

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,144 @@
...
+    val channels = 3
+    val height = image.getHeight
+    val width = image.getWidth
+
+    val decoded = new Array[Byte](height * width * channels)
+    var offset, h = 0
+    while (h < height) {
+      var w = 0
+      while (w < width) {
+        val color = new Color(image.getRGB(w, h))
```
sueann: definitely more readable. we can probably do some array magic with the array version of getRGB if we wanted to try optimizing.

sueann: we could note that, or not, whatever.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,144 @@
...
+   * @return Row image in spark.ml.image format with 3 channels in BGR order.
+   */
+  private[sparkdl] def spImageFromBufferedImage(image: BufferedImage): Row = {
+    val channels = 3
+    val height = image.getHeight
+    val width = image.getWidth
+
+    val decoded = new Array[Byte](height * width * channels)
+    var offset, h = 0
+    while (h < height) {
```
sueann: just curious, are for-loops still slow? i guess they still are: http://dynamicsofprogramming.blogspot.com/2017/01/performance-of-scala-for-loops.html

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,144 @@
...
+
+package com.databricks.sparkdl
+
+import java.awt.image.BufferedImage
+import java.awt.{Color, Image}
+
+import org.apache.spark.image.ImageSchema
+import org.apache.spark.sql.Row
+
+object ImageUtils {
```
sueann: @smurching will this show up in docs because it's not `private[sparkdl]` ?

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -0,0 +1,144 @@
...
+      spImage
+    } else {
+      val srcImg = spImageToBufferedImage(spImage)
+      val tgtImg = new BufferedImage(tgtWidth, tgtHeight, BufferedImage.TYPE_3BYTE_BGR)
+
+      // scaledImg is a java.awt.Image which supports drawing but not pixel lookup by index.
+      val scaledImg = srcImg.getScaledInstance(tgtWidth, tgtHeight, Image.SCALE_AREA_AVERAGING)
+      // Draw scaledImage onto resized (usually smaller) tgtImg so extract individual pixel values.
+      val graphic = tgtImg.createGraphics()
+      graphic.drawImage(scaledImg, 0, 0, null)
```
sueann: looks kinda nuts, but i guess it is what it is ...!

src/test/resources/images
```diff
@@ -0,0 +1 @@
+../../../python/tests/resources/images
```
sueann: can we rename this file to py-images or something that indicates it's pointing somewhere else? was a bit confused by why smaller and bigger images were in different directories.

src/test/scala/com/databricks/sparkdl/ImageUtilsSuite.scala
```diff
@@ -0,0 +1,102 @@
...
+    val tgtWidth: Int = ImageSchema.getWidth(smallerImage)
+    val tgtChannels: Int = ImageSchema.getNChannels(smallerImage)
+
+    val testImage = ImageUtils.resizeImage(tgtHeight, tgtWidth, tgtChannels, biggerImage)
+    assert(ImageSchema.getHeight(testImage) === tgtHeight)
+    assert(ImageSchema.getWidth(testImage) === tgtWidth)
+    assert(ImageSchema.getNChannels(testImage) === tgtChannels)
+    val testImageData = ImageSchema.getData(testImage)
+    val smallerImageData = ImageSchema.getData(smallerImage)
+    assert(testImageData.deep === smallerImageData)
```
sueann: the right side doesn't need `.deep` ? (sorry scala n00b)

src/test/scala/com/databricks/sparkdl/ImageUtilsSuite.scala
```diff
@@ -0,0 +1,102 @@
...
+    val width = 100
+    val channels = 3
+
+    val rand = new Random(971)
+    val imageData = Array.ofDim[Byte](height * width * channels)
+    rand.nextBytes(imageData)
+    val spImage = Row(null, height, width, channels, "CV_U8C3", imageData)
+    val bufferedImage = ImageUtils.spImageToBufferedImage(spImage)
+    val testImage = ImageUtils.spImageFromBufferedImage(bufferedImage)
+    assert(spImage === testImage, "Image changed during conversion.")
```
sueann: why does `"test binary image resize"` above need an element-wise comparison and this okay with comparing the two Row objects?

src/test/scala/com/databricks/sparkdl/ImageUtilsSuite.scala
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  test("Simple BufferedImage from Row Image") {
```
sueann: nit: i'd put this as the first test in this class because it's the most basic case

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
phi-dbq: From our code style guide :D
https://github.com/databricks/scala-style-guide#traversal-and-zipwithindex

sueann: i had hoped the later scala versions would have solved this problem...

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
thunterdb: Probably

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
thunterdb: Let's leave a TODO for now, it is probably not the worst part.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
thunterdb: no, not yet. There are interesting macros that do that, but too much maintenance overhead for now.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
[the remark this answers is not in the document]
thunterdb: Eventually, we should have a case class object for the images, but this is something we can do later.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
thunterdb: interesting...

src/test/scala/com/databricks/sparkdl/ImageUtilsSuite.scala
```diff
@@ -0,0 +1,102 @@
...
+import javax.imageio.ImageIO
+
+import scala.util.Random
+import org.apache.spark.image.ImageSchema
+import org.apache.spark.sql.Row
+import org.scalatest.FunSuite
+
+object ImageUtilsSuite {
+  val biggerImage: Row = {
+    val biggerFile = getClass.getResource("/images/00081101.jpg").getFile
```
thunterdb: can we put that in a subdirectory such as `/images/sparkdl`? Remember that all the namespaces of resources gets flattened into a single namespace within the JVM -> risk of collisions.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
smurching: It'll show up by default, although we can hide `ImageUtils` from the scaladocs if it needs to be exposed so that the Python API can call its methods.

MrBago: I can make it private for now, we can change it later or expose parts of it through a "python interface" type object.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
MrBago: This is the most common way to do it. There is no unique way to map (R, G, B) values to a color people see (you can often mess with the RGB balance on many displays for example) but (x, x, x) maps to a gray/white color in most cases.

src/test/resources/images
MrBago: I moved the images into a sparkdl directory to avoid conflicts, I also renamed the main image directory to "test-image-collection" and the smaller image to "00081101-small-version.png" to make it clear that it's a smaller version of a specific image from the collection.

src/test/scala/com/databricks/sparkdl/ImageUtilsSuite.scala
MrBago: :( this wasn't intentional. I think this works because of scala magic, but I'll add the deep for clarity.

MrBago: Actually it seems like we can compare the spImage Rows directly, hope that's reliable :)

PR #72, opened 2017-10-27
Fix README typo

smurching:
Corrects README to state that we added SQL UDF support for Keras models in DLP 0.2.0 (currently, the README says we added Scala SQL UDF support).


jkbradley:
LGTM

PR #82, opened 2017-11-28
[ML-2851] Add KerasTransformer implementation & tests

smurching:
Adds KerasTransformer, which applies a Tensorflow-backed Keras model (specified by a file name) to a column of arrays in a DataFrame.

This PR includes:
* KerasTransformer implementation & tests
* Changes to `TFTransformer` to support graphs (e.g. Keras model graphs) with input types other than tf.float64.


python/tests/transformers/keras_transformer_test.py
```diff
@@ -0,0 +1,145 @@
...
+from keras.models import Sequential
+from keras.layers import Dense, Activation
+from keras.initializers import glorot_uniform
+
+from sparkdl.transformers.keras_transformer import KerasTransformer
+from ..tests import SparkDLTestCase
+
+class KerasTransformerTest(SparkDLTestCase):
+
+    # TODO(sid): How to create static class variables in Python? Should I just put these in the
```
smurching: How should I handle these constants? Put them in a constructor, make them `@properties`, leave them as is?

python/sparkdl/transformers/keras_transformer.py
```diff
@@ -0,0 +1,64 @@
...
+from sparkdl.param import keyword_only, HasKerasModel, HasInputCol, HasOutputCol
+
+from pyspark.ml import Transformer
+
+from sparkdl.transformers.keras_utils import KSessionWrap
+from sparkdl.graph.input import TFInputGraph
+from .tf_tensor import TFTransformer
+
+
+class KerasTransformer(Transformer, HasInputCol, HasOutputCol, HasKerasModel):
```
smurching: Q: Where should I add a usage example? In the README? Or in the docstring potentially?

python/tests/transformers/keras_transformer_test.py
```diff
@@ -0,0 +1,145 @@
...
+    # constructor?
+    NUM_FEATURES = 10
```
smurching: How should I handle these (static, class-level) constants? Put them in the constructor, make them `@properties`, leave them as is?

python/sparkdl/graph/builder.py
```diff
@@ -207,7 +207,9 @@ def fromList(cls, functions):
         The inputs and outputs are picked out of the scopes, so that users
         will still be able to call the function with the expected inputs/outputs names.
 
-        It is assumed that there is only one input and one output in the intermediary layers
+        It's assumed that the input/output names of adjacent GraphFunction elements in
```
smurching: AFAICT, the updated comment here reflects the function (`GraphFunction.fromList`)'s actual behavior. I'm thinking we should test that this function works when gluing together graphs with > 1 input/output tensor.

IMO the easiest (least code) way to do so would be to add a TFTransformer test with multiple inputs/outputs; the test could also have input types other than tf.float64 (to verify that we now handle other input types).

However I'm also open to creating separate unit tests for this function (verifying gluing together of multiple inputs/outputs) and TFTransformer (verifying that we can handle non-float64 inputs). Open to suggestions :)

python/tests/transformers/keras_transformer_test.py
sueann: @MrBago

python/sparkdl/transformers/keras_transformer.py
sueann: I think the README is better for readability -- we'll create a real guide on the github page you made from the README.

python/sparkdl/graph/builder.py
sueann: Let's add a few tests with very simple graphs for this function specifically. It'll be easier to reason about and detect problems. Thanks sid!

python/tests/tests.py
```diff
@@ -13,7 +13,10 @@
 # limitations under the License.
 #
 
+import shutil
```
smurching: Note: I moved logic for creating/cleaning up a temp directory for storing exported Keras models into the base test class (`SparkDLTestCase`), should I maybe create a new class for this?

python/sparkdl/transformers/tf_tensor.py
```diff
@@ -61,6 +62,49 @@ def setParams(self, tfInputGraph=None, inputMapping=None, outputMapping=None, tf
...
+            # Iterate through the input tensors of the original graph
+            for _, orig_tensor_name in self.getInputMapping():
+                orig_tensor = tfx.get_tensor(orig_tensor_name, user_graph)
+                # Create placeholder with same shape as original input tensor, but that accepts
+                # float64 input from Spark.
+                spark_placeholder = tf.placeholder(tf.float64, shape=orig_tensor.shape,
+                                             name=tfx.op_name(orig_tensor_name))
+                # If the original tensor was of type float64, just pass through the Spark input
+                if orig_tensor.dtype == tf.float64:
+                    output_tensor = tf.identity(spark_placeholder)
```
sueann: do you need the identity op here or can you just use `spark_placeholder` ?

python/sparkdl/transformers/tf_tensor.py
```diff
@@ -69,12 +113,19 @@ def _optimize_for_inference(self):
...
 
+        # Build GraphFunction wrapping the model graph
+        graph_fn = GraphFunction(graph_def=gin.graph_def, input_names=input_node_names,
+                                 output_names=output_node_names)
+        # Inject cast ops to convert float64 input fed from Spark into the datatypes of the
+        # Graph's input nodes.
+        graph_fn_with_casts = self._addCastOps(graph_fn)
+
+        # Optimize for inference, replacing placeholders with float64
         # NOTE(phi-dbq): Spark DataFrame assumes float64 as default floating point type
```
sueann: we can remove this now or move it up into the comment for `_addCastOps`

python/sparkdl/transformers/tf_tensor.py
```diff
@@ -69,12 +113,19 @@ def _optimize_for_inference(self):
...
         output_node_names = [tfx.op_name(tnsr_name) for tnsr_name, _ in output_mapping]
 
+        # Build GraphFunction wrapping the model graph
+        graph_fn = GraphFunction(graph_def=gin.graph_def, input_names=input_node_names,
+                                 output_names=output_node_names)
+        # Inject cast ops to convert float64 input fed from Spark into the datatypes of the
+        # Graph's input nodes.
+        graph_fn_with_casts = self._addCastOps(graph_fn)
+
+        # Optimize for inference, replacing placeholders with float64
```
sueann: do you remember what "optimize for inference" means here? (i forget)

python/tests/transformers/keras_transformer_test.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def _loadNumpyData(self, num_examples, num_features):
+        """
+        Construct and return a 2D numpy array of shape (num_examples, num_features) corresponding
+        to a dataset of one-dimensional examples.
+        """
+        np.random.seed(self.RANDOM_SEED)
+        return np.random.randn(num_examples, num_features)
+
+    def getInputDF(self, sqlContext, inputCol, idCol):
```
sueann: this could also be "private"? or have we been making functions used by `test_` methods "public"? whatever is consistent with other test classes works for me.

python/tests/transformers/keras_transformer_test.py
```diff
@@ -0,0 +1,145 @@
...
+        :param id_col: String, Column assumed to contain a row ID (Long)
+        :param output_col: String, Column containing transform() output. We assume that for
+                           each row, transform() has produced a single-element array containing a
+                           Keras model's prediction on that row)
+        """
+        collected = final_df.collect()
+        collected.sort(key=lambda row: row[id_col])
+        return [row[output_col][0] for row in collected]
+
+    def test_keras_transformer_single_dim(self):
```
sueann: I'd put the test_ methods at the top of the class since they are what matter the most

python/tests/transformers/keras_transformer_test.py
```diff
@@ -0,0 +1,145 @@
...
+        """
+        Compare KerasTransformer output to that of a Keras model on a one-dimensional input dataset.
+        """
+        # Declare column names
```
sueann: no need for a self-explanatory comment :-)

python/tests/transformers/keras_transformer_test.py
```diff
@@ -0,0 +1,145 @@
...
+        # Load dataset, transform it with KerasTransformer
+        df = self.getInputDF(self.sql, inputCol=input_col, idCol=id_col)
+        final_df = transformer.transform(df)
+        sparkdl_predictions = self.transformOutputToComparables(final_df, id_col, output_col)
+
+        # Verify that result DF has the specified input & output columns
+        self.assertDfHasCols(final_df, [input_col, output_col, id_col])
+        self.assertEqual(len(final_df.columns), 3)
+
+        # Compute Keras model output
```
sueann: "Compute Keras model local execution output"

python/tests/transformers/keras_transformer_test.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def prepareKerasModelFile(self, model, filename):
+        """
+        Saves the passed-in keras model to a temporary directory with the specified filename.
+        """
+        path = os.path.join(self.tempdir, filename)
+        model.save(path)
+        return path
+
+    def executeKerasModel(self, df, model, input_col, id_col):
```
sueann: maybe `executeKerasModelLocally` is clearer

python/tests/transformers/keras_transformer_test.py
```diff
@@ -0,0 +1,145 @@
...
+        row ID is contained in id_col).
+        """
+        # Collect dataframe, sort rows by ID column
+        rows = df.select(input_col, id_col).collect()
+        rows.sort(key=lambda row: row[id_col])
+        # Get numpy array (num_examples, num_features) containing input data
+        x_predict = np.array([row[input_col] for row in rows])
+        return model.predict(x_predict)
+
+    def transformOutputToComparables(self, final_df, id_col, output_col):
```
sueann: maybe `convertToKerasOutputFormat` ? "transform" sounds like we're going to call `.transform`

python/tests/transformers/keras_transformer_test.py
sueann: also nit: put below `transformOutputToComparables` so the transformer helper methods are together.

python/tests/tests.py
sueann: yeah probably better to create a new class. also i think it's okay to put it in the same file.

python/sparkdl/transformers/tf_tensor.py
smurching: AFAIK it prunes out ops/nodes that aren't involved in computing the specified output tensors

python/sparkdl/transformers/keras_transformer.py
```diff
@@ -0,0 +1,66 @@
...
+    """
+    Applies a Tensorflow-backed Keras model (specified by a file name) to
+    a column of arrays (where each array corresponds to a Tensor) in a DataFrame.
+    Produces an output column of arrays.
+
+    Restrictions of the current API:
+      * See TFTransformer
+      * Only supports Keras models with a single input tensor & a single output tensor
```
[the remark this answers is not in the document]
smurching: The current implementation technically supports arbitrarily-shaped input/output tensors (where n-dimensional tensors are represented by nested arrays).

However I'm not sure anybody would actually want to use higher-dimensional inputs/outputs (e.g. for images they're better off using KerasImageFileTransformer), & we don't explicitly test higher-dimensional tensors (though they should work).

Maybe we should assert that the input/output are 1D?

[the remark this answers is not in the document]
smurching: chatted offline, i'll add a small test for higher-dim input/output

python/sparkdl/transformers/tf_tensor.py
```diff
@@ -62,91 +65,84 @@ def setParams(self, tfInputGraph=None, inputMapping=None, outputMapping=None, tf
...
-        input_node_names = [tfx.op_name(tnsr_name) for _, tnsr_name in input_mapping]
-        output_node_names = [tfx.op_name(tnsr_name) for tnsr_name, _ in output_mapping]
-
-        # Build GraphFunction wrapping the model graph
-        graph_fn = GraphFunction(graph_def=gin.graph_def, input_names=input_node_names,
-                                 output_names=output_node_names)
         # Inject cast ops to convert float64 input fed from Spark into the datatypes of the
         # Graph's input nodes.
-        graph_fn_with_casts = self._addCastOps(graph_fn)
+        graphdef_with_casts = self._addCastOps(gin.graph_def)
```
sueann: nit: just oneline it since `gin` is not used otherwise:
  graphdef_with_casts = self._addCastOps(self.getTFInputGraph().graph_def)

python/sparkdl/transformers/tf_tensor.py
```diff
@@ -62,91 +65,84 @@ def setParams(self, tfInputGraph=None, inputMapping=None, outputMapping=None, tf
...
+                        input_map[orig_tensor_name] = spark_placeholder
+                    # Otherwise, cast the Spark input to the datatype of the original tensor
+                    else:
+                        input_map[orig_tensor_name] = tf.cast(spark_placeholder,
+                                                              dtype=orig_tensor.dtype)
+            tf.import_graph_def(graph_def=user_graph_def, input_map=input_map, name="")
+        return injected_op_subgraph.as_graph_def(add_shapes=True)
 
     def _optimize_for_inference(self):
         """ Optimize the graph for inference """
```
sueann: not sure this comment helps since it's basically the name of the function - i'd just remove it

python/tests/tests.py
```diff
@@ -54,17 +54,39 @@ def tear_down_env(cls):
...
     @classmethod
     def tearDownClass(cls):
         cls.tear_down_env()
-        shutil.rmtree(cls.tempdir)
 
     def assertDfHasCols(self, df, cols = []):
         map(lambda c: self.assertIn(c, df.columns), cols)
+
+
+class SparkDLTempDirTestCase(SparkDLTestCase, TestTempDir):
```
sueann: I'd normally say we should just have the specific test case classes inherit from both... but since all our test case classes are subclasses of `SparkDLTestCase` we might as well have this...

python/tests/transformers/keras_transformer_test.py
```diff
@@ -16,32 +16,76 @@
...
+        model.add(Activation('sigmoid'))
+        # Compare KerasTransformer output to raw Keras model output
+        self._test_keras_transformer_helper(model, model_filename="keras_transformer_single_dim")
 
-    def getInputDF(self, sqlContext, inputCol, idCol):
-        """ Return a DataFrame containing a long ID column and an input column of arrays. """
-        x_train = self._loadNumpyData(num_examples=20, num_features=self.NUM_FEATURES)
-        train_rows = [{idCol : i, inputCol : x_train[i].tolist()} for i in range(len(x_train))]
-        return sqlContext.createDataFrame(train_rows)
+    def test_keras_transformer_multi_dim(self):
```
sueann: name it `test_keras_transformer_2dim` so it's clear it doesn't work for 3+ dimensions

PR #83, opened 2017-11-29
Adding licensing information for Keras models in code

sueann:
Putting in specific license information for each model used in DeepImageFeaturizer. It essentially says that the models come via Keras, states the Keras license information, and for each model that comes from another source adds the licensing information for the source.


smurching:
This LGTM, lol @ the commit messages. 

I like that all the license info for Python ends up in one file, makes me think it'd be nice to (in the future) move all references to Keras models in Scala into a single file.

PR #86, opened 2017-12-05
Add persistence & test for DeepImageFeaturizer in Scala

smurching:
* Adds persistence for Scala implementation of `DeepImageFeaturizer`. 
  * Since `DeepImageFeaturizer` stores all its data as instances of `org.apache.spark.ml.Param`, all we need to do is extend Spark's default ML persistence traits (`DefaultParamsReadable`, `DefaultParamsWritable`). 
* Adds test for `DeepImageFeaturizer` persistence; we copy over relevant portions of Spark's default ML persistence tests.

See MLlib's [Binarizer](https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/ml/feature/Binarizer.scala) and [BinarizerSuite](https://github.com/apache/spark/blob/branch-2.2/mllib/src/test/scala/org/apache/spark/ml/feature/BinarizerSuite.scala#L107) for an example of a similar Transformer/test suite pair.

### DIFF
--- a/src/main/scala/com/databricks/sparkdl/DeepImageFeaturizer.scala
+++ b/src/main/scala/com/databricks/sparkdl/DeepImageFeaturizer.scala
@@ -23,11 +23,10 @@ import org.apache.spark.ml.Transformer
 import org.apache.spark.ml.linalg.SQLDataTypes.VectorType
 import org.apache.spark.ml.linalg.Vectors
 import org.apache.spark.ml.param.{Param, ParamMap}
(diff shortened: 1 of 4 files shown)


smurching:
As is, this PR enables DeepImageFeaturizer persistence in Scala but not Python; to enable it in Python, we just need to inherit from `JavaMLWriter` & `JavaMLReader`, which define Python persistence methods (`save`, `load`) that call the corresponding Scala methods.

jkbradley:
LGTM

PR #88, opened 2017-12-16
Make python DeepImageFeaturizer use Scala version.

tomasatdatabricks:
- Based of Image schema PR, do not merge until Image schema is merged.
- Otherwise mostly straightforward except results will not match keras in general due to different image libraries


src/main/scala/com/databricks/sparkdl/DeepImageFeaturizer.scala
```diff
@@ -38,6 +38,7 @@ class DeepImageFeaturizer(override val uid: String) extends Transformer with Def
 
   final val inputCol: Param[String] = new Param[String](this, "inputCol", "input column name")
   final val outputCol: Param[String] = new Param[String](this, "outputCol", "output column name")
+  final val scaleFast: Param[Boolean] = new Param[Boolean](this,"scaleFast","use fast resizing if set.")
```
sueann: what does "fast resizing" mean?

python/sparkdl/transformers/named_image.py
```diff
@@ -235,3 +216,45 @@ def _buildTFGraphForName(name, featurize):
     modelData["graph"] = graph
 
     return modelData
+
+class PyDeepImageFeaturizer(Transformer, HasInputCol, HasOutputCol):
```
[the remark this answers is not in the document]
tomasatdatabricks: this is still WIP, I have not decided what to do with Py version of the featurizer. I can mark it as test only.

src/main/scala/com/databricks/sparkdl/DeepImageFeaturizer.scala
tomasatdatabricks: it's just a flag singnaling to the resizer to use the other resize method. Not sure how to name it it's not mapped to a single flag on the java side either. SCALE_FAST or SCALE_DEFAULT would work, probably a few others as well. the name is WIP

src/main/scala/com/databricks/sparkdl/DeepImageFeaturizer.scala
```diff
@@ -147,7 +162,7 @@ object DeepImageFeaturizer extends DefaultParamsReadable[DeepImageFeaturizer] {
 
   // TODO: support batched graphs with mapBlocks
 
-  private[sparkdl] trait NamedImageModel {
+  protected[sparkdl] trait NamedImageModel {
```
MrBago: Why this change? Do we expect this to be changed in subclasses?

src/main/scala/com/databricks/sparkdl/DeepImageFeaturizer.scala
```diff
@@ -95,14 +104,20 @@ class DeepImageFeaturizer(override val uid: String) extends Transformer with Def
     this
   }
 
+  def setResizeFlag(value: String): this.type = {
+    set(scaleHint, value)
```
MrBago: I think we always use the naming convention `setX` for param `X`. Could this be renamed to `setScaleHint` or could we rename the param as `resizeFlag`.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -113,13 +115,15 @@ private[sparkdl] object ImageUtils {
    * @param tgtChannels number of channels of output image (must be 3), may be used later to
    *                    support more channels.
    * @param spImage     image to resize.
+   * @param scaleHint   hint which algorhitm to use, see java.awt.Image#SCALE_DEFAULT
```
MrBago: This suggests that our default is `SCALE_DEFAULT`, but I don't think we do.

MrBago: btw, I think `SCALE_DEFAULT` might be implementation specific for java.awt.Image subclasses.

python/sparkdl/transformers/named_image.py
```diff
@@ -115,48 +120,47 @@ def _getIntermediateOutputCol(self):
...
-        transformer = _NamedImageTransformer(inputCol=self.getInputCol(),
-                                             outputCol=self.getOutputCol(),
-                                             modelName=self.getModelName(), featurize=True)
-        return transformer.transform(dataset)
+        scalaFeaturizer = dataset._sc._jvm.com.databricks.sparkdl.DeepImageFeaturizer()
+        scalaFeaturizer.setModelName(self.getOrDefault(self.modelName))
+        scalaFeaturizer.setInputCol(self.getOrDefault(self.inputCol))
+        scalaFeaturizer.setOutputCol(self.getOrDefault(self.outputCol))
+        if(self.isDefined(self.scaleHint)):
+            scalaFeaturizer.setResizeFlag(self.getOrDefault(self.scaleHint))
```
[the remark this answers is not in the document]
tomasatdatabricks: scaleHint is optional, (input/output)col are not.
The isDefined call is so that we don't override scala's default.

python/sparkdl/utils/generate_app_models.py
```diff
@@ -0,0 +1,156 @@
...
+                    shape=[
+                        1,
+                        model.inputShape()[0],
+                        model.inputShape()[1],
+                        3]))
+            m = model.model(preprocessed=model.preprocess(imageTensor), featurize=featurize)
+            outTensor = tf.to_double(tf.reshape(m.output, [-1]), name="%s_sparkdl_output__" % name)
+            gdef = tfx.strip_and_freeze_until([outTensor], session.graph, session, False)
+            g2 = tf.Graph()
+            with g2.as_default():
```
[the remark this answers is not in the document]
tomasatdatabricks: Yes we do need a new graph into which we read the definition. Otherwise it imports it into the existing graph.

python/sparkdl/utils/generate_app_models.py
```diff
@@ -0,0 +1,156 @@
...
+                        "width": model.inputShape()[1],
+                        "version": version,
+                        "base64": base64_hash})
+                return g2
+
+
+if __name__ == '__main__':
+    x = os.path.abspath(__file__)
+    for i in range(4):
+        x = os.path.split(x)[-2]
```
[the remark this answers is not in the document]
tomasatdatabricks: yeah, good point, remainder of my experimental code when I though split would give me pice wise path

python/sparkdl/utils/generate_app_models.py
```diff
@@ -0,0 +1,156 @@
...
+            '      ModelFetcher.importGraph(Paths.get(file), "jVCEKp1bV53eib8d8OKreTH4fHu/Ji5NHMOsgdVwbMg=")',
+            '        .getOrElse {',
+            '          throw new Exception(s""\"The hash of file $file did not match the expected value.""\".stripMargin)',
+            '        }',
+            '    }',
+            '  }',
+            ''
+        ]))
+        for name, modelConstructor in sorted(
+                keras_applications.KERAS_APPLICATION_MODELS.items(), key=lambda x: x[0]):
```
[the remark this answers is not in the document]
tomasatdatabricks: why?

python/tests/tests.py
```diff
@@ -42,7 +42,7 @@ class PythonUnitTestCase(unittest.TestCase):
 class TestSparkContext(object):
     @classmethod
     def setup_env(cls):
-        cls.sc = SparkContext('local[*]', cls.__name__)
+        cls.sc = SparkContext.getOrCreate()
```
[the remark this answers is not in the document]
tomasatdatabricks: Nope I checked with Philip who is teh author of that line and i is fine this way.
The reason is that I added test which runs prior to this one and which creates a spark context, in which case this test fails with too many spark contexts

src/main/scala/com/databricks/sparkdl/DeepImageFeaturizer.scala
tomasatdatabricks: Yes, I've renamed resizeFlag to scaleHint and forgot to change the method. Good catch

src/main/scala/com/databricks/sparkdl/DeepImageFeaturizer.scala
tomasatdatabricks: I needed to access it from the Models.scala file

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
tomasatdatabricks: hmm I randomly picked one, it is not supposed to suggest what the default value is. I can replace it with default.

python/tests/tests.py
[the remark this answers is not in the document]
phi-dbq: I think right now we run each individual test file separately https://github.com/databricks/spark-deep-learning/blob/master/python/run-tests.sh#L100-L109 in our Travis CI.
(Correct me if I am wrong) I think creating new ones or not shouldn't matter.

tomasatdatabricks:
As for the high level comments:

1) Sure, you might need to change the script. The aim here is not to make script which will work without change forever. It's a script which was used for generating the models/graphs for this version, that's it. If you make changes you might need to change the script. Leaving it in the same repo is mostly for convenience. I really don't see a problem here. I can add comments clarifying this.

2) Regarding the tests, I agree the test run for too long. But are you saying this PR adds significantly to the test runtime? The test seem to be running for pretty much the same time. It think this has been an issue before an so an test . overhaul belongs to a separate PR.

python/tests/tests.py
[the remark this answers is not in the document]
tomasatdatabricks: It matters when you run the tests locally

python/sparkdl/utils/generate_app_models.py
```diff
@@ -0,0 +1,156 @@
...
+        "https://s3-us-west-2.amazonaws.com/spark-deep-learning-models/sparkdl-%(name)s_v%(version)d.pb",
+        fileName = "sparkdl-inceptionV3_v%(version)d.pb",
+        base64Hash = "%(base64)s"
+    )
+  }
+"""
+
+
+def gen_model(name, model, model_file, version=1, featurize=True):
+    g = tf.Graph()
```
sueann: Simpler to read & understand that there is no scope overlap:
```
g = tf.Graph()
with tf.Session(graph=g) as sess:
  ...
  gdef =  tfx.strip_and_freeze_until([outTensor], sess.graph, sess, return_graph=False)
g2 = tf.Graph()
with tf.Session(graph=g2) as sess:    # i believe `with g2.as_default():` instead of initializing a session would also work here but I'd do it outside of the previous session for clarity.
  tf.import_graph_def(gdef, name='')
  filename = "sparkdl-%s_v%d.pb" % (name, version)
  print 'writing out ', filename
  tf.train.write_graph(g2.as_graph_def(), logdir="./", name=filename, as_text=False)
# I'd return here and deal with the scala file writing elsewhere, but if you want to put anything more, you can do it outside the second session.
```

python/sparkdl/transformers/named_image.py
```diff
@@ -115,48 +120,47 @@ def _getIntermediateOutputCol(self):
...
         self._set(**kwargs)
-        return self
-
-    def setModelName(self, value):
-        return self._set(modelName=value)
-
-    def getModelName(self):
-        return self.getOrDefault(self.modelName)
 
     def _transform(self, dataset):
```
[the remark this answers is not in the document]
tomasatdatabricks: I was not aware of JavaTransformer. I'll look into it.

python/sparkdl/utils/generate_app_models.py
tomasatdatabricks: Yeah that's way cleaner, thanks.

python/tests/tests.py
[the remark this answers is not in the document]
phi-dbq: ^ I agree. When testing locally and running multiple tests simultaneously, there will be conflict. What I was trying to say is that I support the way you are changing the spark context.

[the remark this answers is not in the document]
MrBago: Can you guys explain this to me? How are you all running this locally? Each test suite should create and destroy it's spark context. Is that not currently happening?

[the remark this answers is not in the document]
tomasatdatabricks: Ok it might not be dependence between tests. 

named_image.py::_getScaleHintList() calls into jvm to get the list of valid resize hints so it creates a spark context. And it gets called before Philip's setup_env. 

I am not sure what's the proper practice here, should I destroy it after use?

MrBago:
One more thing, I know that we've already talked about style but I don't know if the linter will catch this. We should be consistent about using camelCase as much as possible. And also lets try and be consistent with indentation:

```
scala(
  lines,
  wrap,
  likeThis)

python(wraps, like,
       this)

weShouldNot(use,
            this,
            hybrid,
            style)
```

(We had a conversation about adopting a single indentation style for the whole project back in Jun, and while I think there were some good arguments for that we decided it would best to stick to the predominant style in each of the two languages.)

src/main/scala/com/databricks/sparkdl/DeepImageFeaturizer.scala
MrBago: I think `private[sparkdl]` means you can access it from within the `sparkdl` package.

tomasatdatabricks: True, good catch

python/tests/tests.py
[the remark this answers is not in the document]
MrBago: In general it's not safe to assume that a SparkContext is available on module import. I don't remember why, but I know it's bit us before. That's why we end up with code like this, https://github.com/graphframes/graphframes/pull/169/files#diff-e81e6b169c0aa35012a3263b2f31b330R385

In this case we might be able to avoid using `_getScaleHintList()` if we inherit from `JavaTransformer`.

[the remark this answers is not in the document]
tomasatdatabricks: I don't think inheriting from JavaTransformer changes things here. I still need a Param with validator on the python side.

I guess technically we don't need parameter validator on the python side and just use the one on the scala side. Except then you get the exception only after running transform and not when you set the parameter which is not nice. 

Or I could put initialization into constructor.

tomasatdatabricks:
Ok I've pushed updated version. 

Biggest change - I've changed the gen_app_models.py to not modify the Models.scala directly. 
It now generates the Models.scala.generated file in the working directory and lets user handle the rest.

python/tests/tests.py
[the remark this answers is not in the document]
MrBago: Wait ... does the validator on scala Param not get run until stransform is called? I might look into fixing that after this release.

python/tests/transformers/named_image_test.py
```diff
@@ -184,22 +186,58 @@ def test_prediction_decoded(self):
...
+    def test_featurization_no_reshape(self):
+        """
+        Run sparkDL predictor on manually-resized images and compare result to the
+        keras result.
+        """
+        imageArray = self.imageArray
+        kerasPredict = self.kerasPredict
+
+        def rowWithImage(img):
+            # return [imageIO.imageArrayToStruct(img.astype('uint8'), imageType.sparkMode)]
```
smurching: Remove this line?

python/tests/transformers/named_image_test.py
```diff
@@ -184,22 +186,58 @@ def test_prediction_decoded(self):
...
+            row = imageIO.imageArrayToStruct(img.astype('uint8'))
+            # re-order row to avoid pyspark bug
```
smurching: Optional: If there's a JIRA ticket for the bug note the ticket name (e.g. "SPARK-xxxx") in this comment?

src/main/scala/com/databricks/sparkdl/DeepImageFeaturizer.scala
```diff
@@ -38,6 +41,9 @@ class DeepImageFeaturizer(override val uid: String) extends Transformer with Def
 
   final val inputCol: Param[String] = new Param[String](this, "inputCol", "input column name")
   final val outputCol: Param[String] = new Param[String](this, "outputCol", "output column name")
+  final val scaleHint: Param[String] = new Param(this,"scaleHint","hint which method to use for resizing.",
```
smurching: Nit: Spaces after commas 🙃

src/main/scala/com/databricks/sparkdl/DeepImageFeaturizer.scala
```diff
@@ -95,14 +104,20 @@ class DeepImageFeaturizer(override val uid: String) extends Transformer with Def
     this
   }
 
+  def getScaleHint(value: String): this.type = {
```
smurching: Typo: should this be `setScaleHint` (not `getScaleHint`)?

python/sparkdl/transformers/named_image.py
```diff
@@ -115,48 +120,41 @@ def _getIntermediateOutputCol(self):
...
+    def __init__(self, inputCol=None, outputCol=None, modelName=None, scaleHint=None):
         """
         __init__(self, inputCol=None, outputCol=None, modelName=None)
         """
         super(DeepImageFeaturizer, self).__init__()
         kwargs = self._input_kwargs
-        self.setParams(**kwargs)
-
-    @keyword_only
-    def setParams(self, inputCol=None, outputCol=None, modelName=None):
```
smurching: I think we still need to implement `setParams` along with Python getters/setters for `scaleHint` and `modelName`. You can look at the PySpark's Bucketizer implementation as an example: https://github.com/apache/spark/blob/master/python/pyspark/ml/feature.py#L318

smurching: It's a little strange that our unit tests didn't catch this, in PySpark we run doctests that attempt to get/set params but I guess we don't have equivalent tests in DLP. Would be good to add such tests at some point, probably in a future PR...cc @MrBago

tomasatdatabricks: Ah ok, I did not know this was required. I saw it somewhere in the code but I did not see a use for it. I'll add it, thanks.

python/tests/tests.py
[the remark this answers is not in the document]
tomasatdatabricks: The params are not transferred to scala until you call _transform. At least that how it looks to me after a quick look. I can transfer them eagerly in setParams.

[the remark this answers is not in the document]
tomasatdatabricks: O both work. You can either 
1) leave validation to scala and eagerly transfer params  
or 
2) add converter with lazy initialization.

python/generate_app_models.py
```diff
@@ -0,0 +1,142 @@
+#!/bin/python
```
MrBago: The standard python shebang is `#!/usr/bin/env python`

python/generate_app_models.py
```diff
@@ -0,0 +1,142 @@
...
+
```
MrBago: Can we move this file to `python/scripts/` or `python/model_gen`?

python/sparkdl/transformers/named_image.py
```diff
@@ -115,48 +120,65 @@ def _getIntermediateOutputCol(self):
...
-    The input image column should be 3-channel SpImage.
+    The input image column should be ImageSchema.
     """
 
     modelName = Param(Params._dummy(), "modelName", "A deep learning model name",
                       typeConverter=SparkDLTypeConverters.buildSupportedItemConverter(SUPPORTED_MODELS))
 
+    scaleHint = Param(Params._dummy(), "scaleHint", "Hint which algorhitm to use for image resizing",
+                      typeConverter=_scaleHintConverter)
+
```
MrBago: Did you want to add getter and setter for scaleHint & modelName.

python/sparkdl/transformers/named_image.py
```diff
@@ -115,48 +120,65 @@ def _getIntermediateOutputCol(self):
...
-    def getModelName(self):
-        return self.getOrDefault(self.modelName)
-
-    def _transform(self, dataset):
-        transformer = _NamedImageTransformer(inputCol=self.getInputCol(),
-                                             outputCol=self.getOutputCol(),
-                                             modelName=self.getModelName(), featurize=True)
-        return transformer.transform(dataset)
+# TODO: give an option to take off multiple layers so it can be used in tuning
+#       (could be the name of the layer or int for how many to take off).
```
MrBago: This seems kind of a big task for a source comment, should we maybe track this in an issue or jira instead?

python/tests/graph/test_import.py
```diff
@@ -86,7 +86,8 @@ def test_saved_graph_novar(self):
 
             def gin_fun(session):
                 _build_saved_model(session, saved_model_dir)
-                return TFInputGraph.fromGraph(session.graph, session, [_tensor_input_name], [_tensor_output_name])
+                return TFInputGraph.fromGraph(session.graph, session, [
+                                              _tensor_input_name], [_tensor_output_name])
```
MrBago: style: can we move `[` to next line?

python/tests/graph/test_import.py
```diff
@@ -132,7 +133,8 @@ def test_saved_graph_novar_2(self):
 
             def gin_fun(session):
                 _build_saved_model(session, saved_model_dir)
-                return TFInputGraph.fromGraph(session.graph, session, [_tensor_input_name], [_tensor_output_name])
+                return TFInputGraph.fromGraph(session.graph, session, [
```
MrBago: `[` again.

python/sparkdl/transformers/named_image.py
```diff
@@ -115,48 +120,65 @@ def _getIntermediateOutputCol(self):
...
 
     modelName = Param(Params._dummy(), "modelName", "A deep learning model name",
                       typeConverter=SparkDLTypeConverters.buildSupportedItemConverter(SUPPORTED_MODELS))
 
+    scaleHint = Param(Params._dummy(), "scaleHint", "Hint which algorhitm to use for image resizing",
+                      typeConverter=_scaleHintConverter)
+
     @keyword_only
-    def __init__(self, inputCol=None, outputCol=None, modelName=None):
+    def __init__(self, inputCol=None, outputCol=None, modelName=None, scaleHint=None):
```
MrBago: I think we should set the default value of `scaleHint` to match the default in scala here, in setParams & we need to do `self._setDefault(scaleHint=...)`.

This is a PITA, but I think that's now mllib does it.

src/main/scala/com/databricks/sparkdl/DeepImageFeaturizer.scala
```diff
@@ -18,18 +18,21 @@ package com.databricks.sparkdl
...
+import org.apache.spark.ml.param.{BooleanParam, Param, ParamMap}
 import org.apache.spark.ml.util.{DefaultParamsReadable, DefaultParamsWritable, Identifiable}
-import org.apache.spark.sql.{DataFrame, Dataset, Row}
 import org.apache.spark.sql.functions.{col, udf}
 import org.apache.spark.sql.types.StructType
+import org.apache.spark.sql.{DataFrame, Dataset, Row}
 import org.tensorflow.framework.GraphDef
-import org.tensorframes.{Shape, ShapeDescription}
 import org.tensorframes.impl.DebugRowOps
+import org.tensorframes.{Shape, ShapeDescription}
```
MrBago: Why the re-ordering of all the imports?

src/main/scala/com/databricks/sparkdl/Models.scala
```diff
@@ -0,0 +1,105 @@
+
```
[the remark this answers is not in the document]
MrBago: +1

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -16,9 +16,11 @@
...
 package com.databricks.sparkdl
 
+import java.awt
 import java.awt.image.BufferedImage
 import java.awt.{Color, Image}
 
+import com.sun.javafx.iio.ImageStorage.ImageType
 import org.apache.spark.ml.image.ImageSchema
 import org.apache.spark.sql.Row
 import org.apache.spark.sql.expressions.UserDefinedFunction
```
MrBago: Lets drop the unused imports here:

`awt`, `sql.function`, `sql. expressions` and `com.sum`.

src/main/scala/com/databricks/sparkdl/DeepImageFeaturizer.scala
```diff
@@ -95,14 +104,20 @@ class DeepImageFeaturizer(override val uid: String) extends Transformer with Def
...
   def transform(dataFrame: Dataset[_]): DataFrame = {
     validateSchema(dataFrame.schema)
     val model = DeepImageFeaturizer.supportedModelMap(getModelName)
 
     val imSchema = ImageSchema.columnSchema
     val height = model.height
     val width = model.width
-    val resizeUdf = udf((image: Row) => ImageUtils.resizeImage(height, width, 3, image), imSchema)
+
+    val resizeUdf = udf((image: Row) => ImageUtils.resizeImage(height, width, 3, image, DeepImageFeaturizer.scaleHints(getScaleHint)), imSchema)
```
MrBago: length

python/tests/transformers/named_image_test.py
```diff
@@ -184,22 +185,48 @@ def test_prediction_decoded(self):
...
+        features_sc = np.array([i.features for i in featurizer_sc.transform(
+            self.imageDF).select("features").collect()])
         kerasReshaped = self.kerasFeatures.reshape(self.kerasFeatures.shape[0], -1)
-        np.testing.assert_array_almost_equal(kerasReshaped, features, decimal=6)
+        diffs = [
+            spatial.distance.cosine(
+                kerasReshaped[i],
+                features_sc[i]) for i in range(
+                len(features_sc))]
+        np.testing.assert_array_almost_equal([0 for i in range(len(features_sc))], diffs, decimal=2)
```
MrBago: Numpy lets you compare scalers to arrays:
```
np.testing.assert_array_almost_equal(diffs, 0., decimal=2)
```

python/sparkdl/transformers/named_image.py
tomasatdatabricks: I think the right think to do is not to set the value to anything and let scala side fill in its default. This way guranatees the same default values.

MrBago: Spark actually has a test in the test suite that checks that you've called `_setDefault` with a matching value in python. @jkbradley do you know why we don't delegate defaults to scala for JavaTransformers?

python/tests/tests.py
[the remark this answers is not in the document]
MrBago: I think within mllib `lazy` initialization is mostly used, I'm not sure why that's the case.

src/main/scala/com/databricks/sparkdl/DeepImageFeaturizer.scala
tomasatdatabricks: result Idea's optimize imports. I guess it puts them in alphabetical order.

python/sparkdl/transformers/named_image.py
tomasatdatabricks: Yeah, already added.

src/main/scala/com/databricks/sparkdl/Models.scala
```diff
@@ -0,0 +1,179 @@
...
+    override def graph: GraphDef = ModelFetcher.getFromWeb(
+        "https://s3-us-west-2.amazonaws.com/spark-deep-learning-models/sparkdl-ResNet50_v1.pb",
+        fileName = "sparkdl-inceptionV3_v1.pb",
+        base64Hash = "yrnfcbi84GxJjo30eGkl27GAHxig0JsC106v5wQr3gE="
+    )
+  }
+
+  /**
+    * Model provided by Keras. All cotributions by Keras are provided subject to the
```
sueann: indent

src/main/scala/com/databricks/sparkdl/Models.scala
```diff
@@ -0,0 +1,179 @@
...
+
+    override def graph: GraphDef = ModelFetcher.getFromWeb(
+        "https://s3-us-west-2.amazonaws.com/spark-deep-learning-models/sparkdl-VGG16_v1.pb",
+        fileName = "sparkdl-inceptionV3_v1.pb",
+        base64Hash = "ILVEakOf5arJP11QK4GYDSy8grlUkcfyMR3NqnopXJY="
+    )
+  }
+
+  /**
+    * Model provided by Keras. All cotributions by Keras are provided subject to the
```
sueann: indent

python/sparkdl/transformers/named_image.py
```diff
@@ -115,35 +120,63 @@ def _getIntermediateOutputCol(self):
...
+
     @keyword_only
-    def __init__(self, inputCol=None, outputCol=None, modelName=None):
+    def __init__(self, inputCol=None, outputCol=None, modelName=None, scaleHint="SCALE_AREA_AVERAGING"):
         """
         __init__(self, inputCol=None, outputCol=None, modelName=None)
         """
         super(DeepImageFeaturizer, self).__init__()
+        self._java_obj = self._new_java_obj("com.databricks.sparkdl.DeepImageFeaturizer", self.uid)
+        self._setDefault(scaleHint="SCALE_AREA_AVERAGING")
```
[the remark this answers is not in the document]
tomasatdatabricks: Yeah I am also confused but this seems to be the usage pattern of classes extending JavaTransformer (.e.g see pyspark/ml/feature.py) . I guess the reason is you can reset to the default value by setting the param to None.

src/main/scala/com/databricks/sparkdl/Models.scala
tomasatdatabricks: About the import reordering (git does not let me responded above :/ )

Thanks, the Idea setup helps! I think Idea actually did the right thing in alphabetizing imports from third party libraries. The only violations of the rules is the scala.* reference at the bottom.

python/tests/transformers/named_image_test.py
```diff
@@ -184,22 +185,48 @@ def test_prediction_decoded(self):
...
-        transformed_df = transformer.transform(self.imageDF)
-        collected = self._sortByFileOrder(transformed_df.collect())
-        features = np.array([i.prediction for i in collected])
-
-        # Note: keras features may be multi-dimensional np arrays, but transformer features
-        # will be 1-d vectors. Regardless, the dimensions should add up to the same.
-        self.assertEqual(np.prod(self.kerasFeatures.shape), np.prod(features.shape))
+        # Since we use different libraries for image resizing (PIL in python vs. java.awt.Image in scala),
+        # the result will not match keras exactly. In fact the best we can do is a "somewhat similar" result.
+        # At least compare cosine distance is < 1e-2
```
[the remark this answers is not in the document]
tomasatdatabricks: I think it's fairly meaningful - at least it tests that the vector points to a similar direction. I did comparisons to different images, various amounts of added noise and some obvious bugs I could think of such as  skipping the preprocess or having the color channels flipped. Most distances came orders of magnitude higher, noise with sd = 0.01 got comparable distance. Here's the breakdown on the test images: (the distance metric is by definition from [0,1] interval):

cosine distance per image to the same image with added (normal, mean = 0) noise:
sd = 1.00: [0.69, 0.78, 0.77 0.75, 0.76]
sd = 0.10: [0.1, 0.2, 0.31, 0.12, 0.23]
sd = 0.01: [0.0078, 0.0094, 0.060, 0.0040, 0.0085]

cosine distance with no preprocessing: all  ~ 0.9
cosine distance with faulty preprocess (mean of one channel is incorrect): all ~ 0.1
cosine distance with flipped channels: all ~ 0.3
cosine distance matrix for the test images:
 [ 0.00   0.75   0.71   0.70   0.67]
 [ 0.75   0.00   0.79   0.85   0.74]
 [ 0.71   0.80   0.00   0.75   0.69]
 [ 0.70   0.85   0.75   0.00   0.70]
 [ 0.67   0.74   0.69   0.70   0.00]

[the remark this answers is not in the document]
tomasatdatabricks: I can add a comment about the transferred learning one-time test run but I don't think that's wise unless we actually turn it into a test (the comment could become obsolete after some more changes but not removed). 

Overall I think the combination of (1) you get exact match with no resize and (2) not too different with resize is good enough.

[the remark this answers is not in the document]
sueann: OK I think having your note above in the PR is good enough for documentation - we could maybe put it in the commit msg for better discoverability.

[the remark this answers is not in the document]
tomasatdatabricks: Ok, I'll put it in the commit message as well.

python/model_gen/generate_app_models.py
```diff
@@ -81,60 +86,155 @@ def gen_model(name, model, model_file, version=1, featurize=True):
...
-    model_file.write(
+    model_file.write(indent(
         scala_template % {
+            "license": license,
             "name": name,
             "height": model.inputShape()[0],
             "width": model.inputShape()[1],
             "version": version,
-            "base64": base64_hash})
+            "base64": base64_hash},2))
```
sueann: space after ,

python/model_gen/generate_app_models.py
```diff
@@ -81,60 +86,155 @@ def gen_model(name, model, model_file, version=1, featurize=True):
...
-            ''
-        ]))
+        f.write(models_scala_header)
         for name, modelConstructor in sorted(
                 keras_applications.KERAS_APPLICATION_MODELS.items(), key=lambda x: x[0]):
             print 'generating model', name
-            g = gen_model(name=name, model=modelConstructor(), model_file=f)
+            if not name in licenses:
+                raise KeyError("Missing license for model '%s'" % name )
+            g = gen_model(license = licenses[name],name=name, model=modelConstructor(), model_file=f)
```
sueann: space after ,

python/sparkdl/transformers/named_image.py
```diff
@@ -159,20 +157,19 @@ class DeepImageFeaturizer(JavaTransformer, HasInputCol, HasOutputCol):
...
         self.setParams(**kwargs)
 
-
     @keyword_only
-    def setParams(self, inputCol=None, outputCol=None, modelName=None, scaleHint=None):
+    def setParams(self, inputCol=None, outputCol=None, modelName=None, scaleHint="SCALE_AREA_AVERAGING"):
         """
         setParams(self, inputCol=None, outputCol=None, modelName=None, decodePredictions=False,
-                  topK=5)
+                  scaleHint="SCALE_AREA_AVERAGING",topK=5)
```
sueann: space after ,

python/model_gen/generate_app_models.py
```diff
@@ -37,24 +39,27 @@
...
+      fileName = "sparkdl-inceptionV3_v%(version)d.pb",
+      base64Hash = "%(base64)s"
+  )
+}
 """
 
 
-def gen_model(name, model, model_file, version=1, featurize=True):
+def indent(s, lvl):
+    return '\n'.join([' '*lvl + x for x in s.split('\n')])
```
sueann: space around *

python/model_gen/generate_app_models.py
sueann: i think this might need to be 1 instead of 2

sueann:
oh and need to rerun travis

python/model_gen/generate_app_models.py
tomasatdatabricks: I think 2 is correct, it's the number of spaces for the indent, not a number of indent levels.

python/sparkdl/transformers/named_image.py
```diff
@@ -115,35 +118,62 @@ def _getIntermediateOutputCol(self):
...
+        self._setDefault(scaleHint="SCALE_AREA_AVERAGING")
         kwargs = self._input_kwargs
         self.setParams(**kwargs)
 
     @keyword_only
-    def setParams(self, inputCol=None, outputCol=None, modelName=None):
+    def setParams(self, inputCol=None, outputCol=None, modelName=None, scaleHint="SCALE_AREA_AVERAGING"):
         """
-        setParams(self, inputCol=None, outputCol=None, modelName=None)
+        setParams(self, inputCol=None, outputCol=None, modelName=None, decodePredictions=False,
```
sueann: remove decodePredictions & topK (not sure if we talked about this - the comment here goes to docs).

src/main/scala/com/databricks/sparkdl/DeepImageFeaturizer.scala
```diff
@@ -38,6 +39,9 @@ class DeepImageFeaturizer(override val uid: String) extends Transformer with Def
 
   final val inputCol: Param[String] = new Param[String](this, "inputCol", "input column name")
   final val outputCol: Param[String] = new Param[String](this, "outputCol", "output column name")
+  final val scaleHint: Param[String] = new Param(this,"scaleHint", "hint which method to use for resizing.",
```
sueann: space after `this,`

python/model_gen/generate_app_models.py
sueann: Mm not sure what's going on but the generated file's indent doesn't look right:
```
/**
  * What's in Models.scala
  */
```
```
/**
 * Expected
 */
```

python/tests/transformers/named_image_test.py
```diff
@@ -16,6 +16,8 @@
 import numpy as np
 import os
 
+from scipy import spatial
```
sueann: what's going on with the import ordering & grouping here 😂 (lines 16-25)

python/sparkdl/transformers/named_image.py
tomasatdatabricks: oh yeah, that's a good catch, thanks

python/sparkdl/transformers/named_image.py
```diff
@@ -115,35 +118,61 @@ def _getIntermediateOutputCol(self):
...
+        self._java_obj = self._new_java_obj("com.databricks.sparkdl.DeepImageFeaturizer", self.uid)
+        self._setDefault(scaleHint="SCALE_AREA_AVERAGING")
         kwargs = self._input_kwargs
         self.setParams(**kwargs)
 
     @keyword_only
-    def setParams(self, inputCol=None, outputCol=None, modelName=None):
+    def setParams(self, inputCol=None, outputCol=None, modelName=None, scaleHint="SCALE_AREA_AVERAGING"):
         """
         setParams(self, inputCol=None, outputCol=None, modelName=None)
```
sueann: add `scaleHint="SCALE_AREA_AVERAGING"`

PR #90, opened 2017-12-28
Move sparkdl utilities for conversion between numpy arrays and image schema to ImageSchema

tomasatdatabricks:
[WIP] Preparation for moving stuff to Spark. 

Moved utilities for image schema <=> numpy array conversion to (copy pasted from spark 2.3) Image schema code.

1. Extended ImageSchema scala code with support/information for all OpenCv modes
2. python toNDArray and toImage utilities extended to work with all supported data types.
3. [minor] sparkdl toImage function included batch size stripping - had to make a separate call for that


python/sparkdl/image/image.py
```diff
@@ -126,15 +161,20 @@ def toNDArray(self, image):
...
+                "Unexpected number of channels, image has %d channels but OcvType '%s' expects %d channels." %
+                (nChannels, ocvType.name, ocvType.nChannels))
+        itemSz = ocvType.nptype.itemsize
         return np.ndarray(
             shape=(height, width, nChannels),
-            dtype=np.uint8,
+            dtype=ocvType.nptype,
             buffer=image.data,
-            strides=(width * nChannels, nChannels, 1))
+            strides=(width * nChannels * itemSz, nChannels * itemSz, itemSz))
```
MrBago: Will numpy figure out the right strides if we don't pass it explicitly?

python/sparkdl/image/image.py
```diff
@@ -152,29 +192,29 @@ def toImage(self, array, origin=""):
                 "array argument should be numpy.ndarray; however, it got [%s]." % type(array))
 
         if array.ndim != 3:
-            raise ValueError("Invalid array shape")
+            raise ValueError("Invalid array shape %s" % str(array.shape))
```
MrBago: Do we want to reshape 2d arrays to be `shape + (1,)`?

python/sparkdl/image/image.py
```diff
@@ -152,29 +192,29 @@ def toImage(self, array, origin=""):
...
-            mode = ocvTypes["CV_8UC4"]
-        else:
-            raise ValueError("Invalid number of channels")
+        dtype = array.dtype
+        if not dtype in self._numpyToOcvMap:
+            raise ValueError(
+                "Unexpected/unsupported array data type '%s', currently only supported formats are %s" %
+                (str(
+                    array.dtype), str(
+                    self._numpyToOcvMap.keys())))
```
MrBago: Can we get this on fewer lines or sue some variables, it looks odd.

src/test/scala/com/databricks/sparkdl/DeepImageFeaturizerSuite.scala
```diff
@@ -128,4 +128,6 @@ class DeepImageFeaturizerSuite extends FunSuite with TestSparkContext with Defau
       .setOutputCol("myOutput")
     testDefaultReadWrite(featurizer)
   }
+
+
```
MrBago: Extra white space.

python/sparkdl/image/image.py
```diff
@@ -71,11 +87,30 @@ def ocvTypes(self):
...
-            self._ocvTypes = dict(ctx._jvm.org.apache.spark.ml.image.ImageSchema.javaOcvTypes())
-        return self._ocvTypes
+            ctx = SparkContext.getOrCreate()
+            self._ocvTypes = [self._OcvType(name=x.name(),
+                                            mode=x.mode(),
+                                            nChannels=x.nChannels(),
+                                            dataType=x.dataType(),
+                                            nptype=self._ocvToNumpyMap[x.dataType()])
+                              for x in ctx._jvm.org.apache.spark.ml.image.ImageSchema.javaOcvTypes()]
+        return [x for x in self._ocvTypes]
```
MrBago: What does this do? Isn't `self._ocvTypes` already a list?

python/sparkdl/image/image.py
tomasatdatabricks: Hmm yeah I would think so. The original code from ms folks was like this and I did not want to do more changes than necessary.

python/sparkdl/image/image.py
tomasatdatabricks: I agree with their approach. I think it's better to make the caller pass the arguments in expected format rather than trying to auto-convert unless that is completely unambiguous. 

So in this case, we say images are always 3 dimensional arrays and it's up to the user to make sure they conform to that. Otherwise they might be passing something else than they think they are passing and we would mask their bug until later.

python/sparkdl/image/image.py
tomasatdatabricks: Yeah it does, I think it's the autopep8 being weird here. I'll reformat that.

python/sparkdl/image/image.py
tomasatdatabricks: The purpose was to return copy of the list so that the private member can not be modified.

MrBago: oic, I usually see `myList[:]` or `list(myList)` to make a copy.

tomasatdatabricks: Thanks, that's nicer :) 
The members of the list are tuples so shallow copy suffices here.

tomasatdatabricks:
@sueann Yes I agree, I would merge spark version first and merge this one only after spark 2.4 is released. I made the PR here mostly because that's what we need the changes for, so it can be reviewed in context, also to run tests.  

I'll mark it WIP.

sueann:
ah ok got it. thanks!

PR #94, opened 2018-01-24
Update ImageUtils to support resizing one, three, or four channel images

smurching:
This PR:
* Updates conversions from row image to/from Buffered image (`spImageToBufferedImage` and `spImageFromBufferedImage`) to support one, three, and four channel images
* Updates resizeImage to use the `tgtChannels` parameter to determine the number of channels in the output image instead of defaulting to three output channels
* Updates existing tests to verify that resizing, conversions to/from BufferedImage work for one, three, and four-channel images


src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -72,15 +90,40 @@ private[sparkdl] object ImageUtils {
     image
   }
 
+  /** Returns the number of channels in the passed-in buffered image. */
```
smurching: This method and `getOCVType` invert the conversion done in [ImageSchema](https://github.com/apache/spark/blob/1edb3175d8358c2f6bfc84a0d958342bd5337a62/mllib/src/main/scala/org/apache/spark/ml/image/ImageSchema.scala#L146)

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -89,27 +132,37 @@ private[sparkdl] object ImageUtils {
...
       var w = 0
       while (w < width) {
-        val color = new Color(image.getRGB(w, h))
-        decoded(offset) = color.getBlue.toByte
-        decoded(offset + 1) = color.getGreen.toByte
-        decoded(offset + 2) = color.getRed.toByte
+        val color = new Color(image.getRGB(w, h), image.getColorModel.hasAlpha)
+        channels match {
+          case 1 =>
+            decoded(offset) = color.getBlue.toByte
```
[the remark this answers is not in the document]
smurching: Actually I think this can be fixed by stealing some code from ImageSchema, will push an update...

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -42,74 +44,127 @@ private[sparkdl] object ImageUtils {
...
+    }
 
-        val color = new Color(r & 0xff, g & 0xff, b & 0xff)
-        image.setRGB(w, h, color.getRGB)
-        offset += channels
-        w += 1
+    val isGray = image.getColorModel.getColorSpace.getType == ColorSpace.TYPE_GRAY
+    val hasAlpha = image.getColorModel.hasAlpha
+    var offset = 0
+    // Grayscale images in Java require special handling to get the correct intensity
```
smurching: The logic here is basically the inverse of Spark's logic for reading Grayscale images into a data array (this code sets the pixels of a grayscale image from a data array): https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/ml/image/ImageSchema.scala#L159

For context, if we set grayscale image data using `image.setRGB(w, h, rgbInt)`, calling `image.getRGB(w, h)` won't necessarily return the same result (see [this gist](https://gist.github.com/smurching/d6e918a84c79155b616f9339b30fdfca)). Hence we set grayscale image data directly in the raster here.

smurching:
Tests failed for Python 3.5 because curling travis failed, retriggering the tests...

src/test/scala/com/databricks/sparkdl/ImageUtilsSuite.scala
```diff
@@ -22,54 +22,58 @@ import javax.imageio.ImageIO
...
+    // Helper for getting filenames of pre-resized, resized images with different numbers of color
+    // channels
+    def getImagePath(imageSize: String, numChannels: Int): String = {
+      s"/sparkdl/test-image-collection/${numChannels}_channels/$imageSize.png"
+    }
 
-    val testImage = ImageUtils.resizeImage(tgtHeight, tgtWidth, tgtChannels, biggerImage)
-    assert(testImage === smallerImage, "Resizing image did not produce expected smaller image.")
+    for (channels <- Seq(1, 3, 4)) {
+      val smallerImage = ImageUtilsSuite.getImageRow(getImagePath("small", channels))
```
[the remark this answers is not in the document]
tomasatdatabricks: Do you mean you compare against image resized using BufferedImage or  ImageUtils.resizeImage?  

Just a suggestion, why don't we store only the original image, and compare against the image resized in java using BufferedImage on the fly?

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -130,7 +183,12 @@ private[sparkdl] object ImageUtils {
       spImage
     } else {
       val srcImg = spImageToBufferedImage(spImage)
-      val tgtImg = new BufferedImage(tgtWidth, tgtHeight, BufferedImage.TYPE_3BYTE_BGR)
+      val tgtImgType = tgtChannels match {
+        case 1 => BufferedImage.TYPE_BYTE_GRAY
+        case 3 => BufferedImage.TYPE_3BYTE_BGR
```
tomasatdatabricks: I would add 
case _  => throw meaningful exception.

smurching:
Just pushed an update that hopefully addresses the ^comments :)

> The cheating test should be fixed (if I understood the comment :)
Addressed by adding a helper method that performs an image resize using as few spark-deep-learning utilities as possible (we still call into spImageFromBufferedImage) and comparing ImageUtils.resizeImage output to the result of the helper.

> I think we should check that the image is one of the supported OpenCvTypes and throw meaningful exception otherwise.
Done

> I would consider using the direct raster access for both grayscale and colored images.
Updated the `spImagetoBufferedImage`, `spImagetoBufferedImage` methods to use the raster code path

Also another clarifying comment: the test images 1_channels/00074201.png, 3_channels/00074201.png, and 4_channels/00074201.png are variants of the three-channel 00074201.jpg image. The alpha channel of 4_channels/00074201.png is set to random values between 122 and 255.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -35,81 +47,88 @@ private[sparkdl] object ImageUtils {
...
-            r = b
-          case 3 =>
-            b = imageData(offset)
-            g = imageData(offset + 1)
-            r = imageData(offset + 2)
-          case _ =>
-            require(false, s"`Channels` must be 1 or 3, got $channels.")
-        }
+    val imageType = openCVModeToImageType.getOrElse(mode,
+      throw new UnsupportedOperationException("Cannot convert row image with  " +
```
smurching: @tomasatdatabricks As of right now, this error can only occur if users try to pass an image row with an unsupported OpenCV mode to Scala's DeepImageFeaturizer. I wonder if the current error message is vague/unhelpful to the user in this case...maybe the image validation should be pushed into DeepImageFeaturizer?

tomasatdatabricks: Yes I think it should definitely be here as well. I believe you should always try to throw informative exception.   Deep ImageFeaturizer is just one of the possible users of this function. Also, it does not have to have the same restrictions.

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -35,81 +47,89 @@ private[sparkdl] object ImageUtils {
...
+          val cIdx = channels - c
+          decoded(offset) = raster.getSample(w, h, cIdx).toByte
+          offset += 1
+        }
       }
-      h += 1
     }
-    Row(origin, height, width, channels, ImageSchema.ocvTypes("CV_8UC3"), decoded)
+    Row(origin, height, width, channels, getOCVType(image), decoded)
+
```
tomasatdatabricks: Nit: extra blank line

src/main/scala/com/databricks/sparkdl/ImageUtils.scala
```diff
@@ -35,81 +47,88 @@ private[sparkdl] object ImageUtils {
...
+        s"${supportedModes.map(ImageSchema.ocvTypes(_)).mkString(", ")}"))
 
-        val color = new Color(r & 0xff, g & 0xff, b & 0xff)
-        image.setRGB(w, h, color.getRGB)
-        offset += channels
-        w += 1
+    val image = new BufferedImage(width, height, imageType)
+    var offset = 0
+    val raster = image.getRaster
+    // NOTE: This code assumes the raw image data in rowImage directly corresponds to the
```
sueann: Does this change the data compared to the previous way?

smurching: Good Q, it doesn't change the data layout compared to the previous way. The BufferedImage types supported by our utilities (`TYPE_BYTE_GRAY, TYPE_3_BYTE_BGR, TYPE_4_BYTE_AGBR`) have a fixed BGR(A) channel ordering so this approach should also be consistent across Java versions etc (see [BufferedImage docs](https://docs.oracle.com/javase/7/docs/api/java/awt/image/BufferedImage.html)).

sueann: Ah I was wondering more about the content, whether what we're doing manually produced the exact same result as `Color` & `Color.getRGB`. The tests look like they maintain the same resizing results as before the code change here but wanted to confirm, since doing anything different to the image can affect the results of applying models to it. (Part of it is, I'm not sure how reasonable this assumption stated here is.)

smurching: Ahh I see. Actually the resulting BufferedImage may change when feeding grayscale images to DeepImageFeaturizer, but not for three-channel images (our DIF tests only run against three-channel images so they don't catch the change).

The difference for grayscale images is that running `val b = rowImageData(offset); val color = new Color(b, b, b); image.setRGB(x, y, color.getRGB)` as the original code did doesn't necessarily set the grayscale pixel at `(x, y)` to `b`. This happens because Java takes a weighted combination of the (r, g, b) channels when setting the grayscale pixel value. However with the raster approach the pixel at `(x, y)` is actually directly set to `b`. Given that `b` was likely obtained by calling `raster.getSample` (how `ImageSchema.readImages` reads the data of a grayscale image), this seems ok to me.

That being said I could see changes in model behavior being sufficiently undesirable that we wouldn't want to make this change.

Another thing: just realized we no longer validate that the input images to DIF are either 1 or 3 channels, I'm assuming we should add this validation back? The practical implication being that as it stands, this PR allows users to feed 4-channel images to DIF without hitting an error.

smurching: @tomasatdatabricks mainly I just didn't want to inadvertently release support for four-channel inputs in DIF without adding unit tests :P. Although I suppose we already supported one-channel inputs without specifically testing DIF on one-channel images.

sueann: ok the current logic looks okay to me. the most important thing is not to change the logic if possible so there is no breaking change for users. i doubt this is a problem at this point. 

for the one- and four- channel input stuff -- if we want any testing, it seems enough to have unit tests to make sure this resizing function returns the correct number of channels given `targetChannels`. Just checking the number of channels seems good enough - we could also check the content of the image if we want to catch any breaking changes in the future.

smurching: SGTM. Should we add the unit tests in this PR or in a follow-up?

sueann:
@smurching I'm happy to merge this. You can add the tests in this PR if it's useful for your Databricks PR, but can delay to another PR if it's not.

smurching:
@sueann thanks, let's go ahead & merge this and add unit tests in a future PR :)

PR #95, opened 2018-01-24
Persistence for DeepImageFeaturizer

tomasatdatabricks:
Inherit MLWriteable and MLReadable plus a bit of extra code to make it work (MLLib code has some hardcoded paths which don't work for deep learning pipelines.

### DIFF
--- a/python/sparkdl/transformers/named_image.py
+++ b/python/sparkdl/transformers/named_image.py
@@ -157,7 +157,11 @@ class DeepImageFeaturizer(JavaTransformer, JavaMLReadable, JavaMLWritable):
     can be used in a MLlib Pipeline.
     The input image column should be ImageSchema.
     """
-
+    # NOTE: We are not inheriting from HasInput/HasOutput to mirror the scala side. It also helps
+    # us to avoid issue with serialization/deserialization - default values based on uid do not
+    # always get set to correct value on the python side after deserialization. Default values do
+    # not get reset to the jvm side value unless they param value is not set.
+    # See pyspark.ml.wrapper.JavaParams._transfer_params_from_java
     inputCol = Param(Params._dummy(), "inputCol", "input column name.", typeConverter=TypeConverters.toString)
     outputCol = Param(Params._dummy(), "outputCol", "output column name.", typeConverter=TypeConverters.toString)
     modelName = Param(Params._dummy(), "modelName", "A deep learning model name",
@@ -205,7 +209,7 @@ class DeepImageFeaturizer(JavaTransformer, JavaMLReadable, JavaMLWritable):
         return self.getOrDefault(self.modelName)
 
     def setScaleHint(self, value):
-            return self._set(scaleHint=value)
+        return self._set(scaleHint=value)
 
     def getScaleHint(self):
         return self.getOrDefault(self.scaleHint)
--- a/python/sparkdl/transformers/named_image.py
+++ b/python/sparkdl/transformers/named_image.py
@@ -20,6 +20,7 @@ import py4j
 from pyspark import SparkContext
 from pyspark.ml import Transformer
 from pyspark.ml.param import Param, Params, TypeConverters
+from pyspark.ml.util import JavaMLReadable, JavaMLWritable, JavaMLReader
 from pyspark.ml.wrapper import JavaTransformer
 from pyspark.sql.functions import udf
 from pyspark.sql.types import (ArrayType, FloatType, StringType, StructField, StructType)
@@ -140,7 +141,16 @@ class _LazyScaleHintConverter:
 _scaleHintConverter = _LazyScaleHintConverter()
 
 
-class DeepImageFeaturizer(JavaTransformer, HasInputCol, HasOutputCol):
+class _DeepImageFeaturizerReader(JavaMLReader):
+    def __init__(self, clazz):
+        super(_DeepImageFeaturizerReader, self).__init__(clazz)
+
+    @classmethod
+    def _java_loader_class(cls, clazz):
+        return "com.databricks.sparkdl.DeepImageFeaturizer"
+
+
+class DeepImageFeaturizer(JavaTransformer, JavaMLReadable, JavaMLWritable, HasInputCol, HasOutputCol):
     """
     Applies the model specified by its popular name, with its prediction layer(s) chopped off,
     to the image column in DataFrame. The output is a MLlib Vector so that DeepImageFeaturizer
@@ -161,6 +171,7 @@ class DeepImageFeaturizer(JavaTransformer, HasInputCol, HasOutputCol):
         """
         super(DeepImageFeaturizer, self).__init__()
         self._java_obj = self._new_java_obj("com.databricks.sparkdl.DeepImageFeaturizer", self.uid)
+        self._defaultParamMap = {} # discard unwanted defaults from parents
         self._setDefault(scaleHint="SCALE_AREA_AVERAGING")
         kwargs = self._input_kwargs
         self.setParams(**kwargs)
@@ -186,6 +197,22 @@ class DeepImageFeaturizer(JavaTransformer, HasInputCol, HasOutputCol):
 
     def getScaleHint(self):
         return self.getOrDefault(self.scaleHint)
+
+    @classmethod
+    def read(cls):
+        return _DeepImageFeaturizerReader(cls)
+
+    @classmethod
+    def _from_java(cls, java_stage):
+        """
+        Given a Java object, create and return a Python wrapper of it.
+        Used for ML persistence.
+        """
+        res = DeepImageFeaturizer()
+        res._java_obj = java_stage
+        res._resetUid(java_stage.uid())
+        res._transfer_params_from_java()
+        return res
 
 
 class _NamedImageTransformer(Transformer, HasInputCol, HasOutputCol):
--- a/python/sparkdl/transformers/named_image.py
+++ b/python/sparkdl/transformers/named_image.py
@@ -150,7 +150,7 @@ class _DeepImageFeaturizerReader(JavaMLReader):
         return "com.databricks.sparkdl.DeepImageFeaturizer"
 
 
-class DeepImageFeaturizer(JavaTransformer, JavaMLReadable, JavaMLWritable, HasInputCol, HasOutputCol):
+class DeepImageFeaturizer(JavaTransformer, JavaMLReadable, JavaMLWritable):
     """
     Applies the model specified by its popular name, with its prediction layer(s) chopped off,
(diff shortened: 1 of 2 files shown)


smurching:
Thanks @tomasatdatabricks, looking at this now

python/sparkdl/transformers/named_image.py
```diff
@@ -187,6 +199,23 @@ def setScaleHint(self, value):
     def getScaleHint(self):
         return self.getOrDefault(self.scaleHint)
 
+    @classmethod
```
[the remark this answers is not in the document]
smurching: ^That being said I'm wondering if it's worth it to have the Python persistence call into Scala via internal APIs, or if we should just call into Python persistence (i.e. extend DefaultParamsReadable & DefaultParamsWritable in DIF, see [this blog post](https://databricks.com/blog/2017/08/30/developing-custom-machine-learning-algorithms-in-pyspark.html)). I can't find any Spark tests verifying that Python persistence produces the same results as Scala persistence, so maybe the current approach is best for correctness.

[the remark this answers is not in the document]
tomasatdatabricks: 1) @smurching  your analysis is accurate, thanks for writing it down!
Similar situation occurs in the other direction when instantiating the Python object - hence the need to override  _from_java method.

2) That's a good question. I decided to do it this way to be as close to other spark examples as possible. Also I wanted the python wrapper to be as thin as possible.

python/sparkdl/transformers/named_image.py
```diff
@@ -140,7 +141,16 @@ def __call__(self, value):
...
 
 
-class DeepImageFeaturizer(JavaTransformer, HasInputCol, HasOutputCol):
+class _DeepImageFeaturizerReader(JavaMLReader):
+    def __init__(self, clazz):
+        super(_DeepImageFeaturizerReader, self).__init__(clazz)
+
+    @classmethod
+    def _java_loader_class(cls, clazz):
+        return "com.databricks.sparkdl.DeepImageFeaturizer"
```
sueann: with the default implementation, is the problem that we don't get `com.databricks.sparkdl`, just `sparkdl` ?

python/tests/transformers/named_image_test.py
```diff
@@ -240,3 +242,31 @@ def test_featurizer_in_pipeline(self):
...
+                                          outputCol="features0", scaleHint="SCALE_FAST")
+        path = tempfile.mkdtemp()
+        dst_path = path + "/featurizer"
+        transformer0.save(dst_path)
+        transformer1 = DeepImageFeaturizer.load(dst_path)
+        self.assertEqual(transformer0.uid, transformer1.uid)
+        self.assertEqual(type(transformer0.uid), type(transformer1.uid))
+        self.assertEqual(transformer1.uid, transformer1.scaleHint.parent,
+                         "Loaded DeepImageFeaturizer instance uid (%s) did not match Param's uid (%s)"
+                         % (transformer1.uid, transformer1.scaleHint.parent))
```
sueann: I'd check some other param since `scaleHint` has a default value. This test doesn't make sure other params are correctly read in.

python/sparkdl/transformers/named_image.py
```diff
@@ -187,6 +198,22 @@ def setScaleHint(self, value):
     def getScaleHint(self):
         return self.getOrDefault(self.scaleHint)
 
+    @classmethod
+    def read(cls):
+        return _DeepImageFeaturizerReader(cls)
+
+    @classmethod
+    def _from_java(cls, java_stage):
```
sueann: Is https://github.com/apache/spark/blob/41bb1ddc63298c004bb6a6bb6fff9fd4f6e44792/python/pyspark/ml/wrapper.py#L218 the reason we can't use the default implementation? (e.g. if the class loader logic was in a separate method we could just overwrite that?)

python/sparkdl/transformers/named_image.py
```diff
@@ -161,6 +171,7 @@ def __init__(self, inputCol=None, outputCol=None, modelName=None, scaleHint="SCA
         """
         super(DeepImageFeaturizer, self).__init__()
         self._java_obj = self._new_java_obj("com.databricks.sparkdl.DeepImageFeaturizer", self.uid)
+        self._defaultParamMap = {} # discard unwanted defaults from parents
```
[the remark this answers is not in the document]
tomasatdatabricks: There is a default from HasOutput which we do not want (the scala side has no default at least).

python/tests/transformers/named_image_test.py
tomasatdatabricks: Sure, I can check uids on other params as well, but I don't think default value changes uid. As for the value, I check the whole paramMap.

python/tests/transformers/named_image_test.py
```diff
@@ -240,3 +242,31 @@ def test_featurizer_in_pipeline(self):
...
+        self.assertEqual(transformer0._paramMap, transformer1._paramMap,
+                         "Loaded DeepImageFeaturizer instance params (%s) did not match "
+                         % str(transformer1._defaultParamMap) +
```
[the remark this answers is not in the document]
tomasatdatabricks: yes

python/tests/transformers/named_image_test.py
```diff
@@ -240,3 +242,31 @@ def test_featurizer_in_pipeline(self):
...
+                         "original defaults (%s)" % str(transformer0._defaultParamMap))
```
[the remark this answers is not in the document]
tomasatdatabricks: yes as well

python/sparkdl/transformers/named_image.py
tomasatdatabricks: yes.

python/sparkdl/transformers/named_image.py
tomasatdatabricks: We get sparkdl.transformers.named_image.DeepImageFeaturizer, the paths are almost completely different.

python/tests/transformers/named_image_test.py
```diff
@@ -240,3 +242,31 @@ def test_featurizer_in_pipeline(self):
         pred_df_collected = lrModel.transform(train_df).collect()
         for row in pred_df_collected:
             self.assertEqual(int(row.prediction), row.label)
+
+
+class DeepImageFeaturizerPersistenceTest(SparkDLTestCase):
+    def test_inception(self):
+        transformer0 = DeepImageFeaturizer(inputCol='image', modelName="InceptionV3",
+                                          outputCol="features0", scaleHint="SCALE_FAST")
+        path = tempfile.mkdtemp()
```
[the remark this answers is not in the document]
tomasatdatabricks: Thanks, that's a good point.

tomasatdatabricks:
Thanks for the reviews @sid @sueann! Yeah I was also surprised the default implementation did not work and there was no easy way how to make it work. E.g. I think making the load method a method instead of class method and including a member variable with the java name would help but I did not want to get too deep into it. I would love to hear Joseph's feedback on it when he gets back.

python/sparkdl/transformers/named_image.py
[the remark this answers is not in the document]
sueann: this seems like a bug waiting to happen (person subclassing / taking a trait will expect to inherit the default values). do you think this needs to be overwritten generally, or just for the persistence (read) case? if the latter, we can handle this specific case in `_from_java` before `res._transfer_params_from_java()` is called.

(ugh it's weird / annoying that the scala version doesn't have a default value and the python version does)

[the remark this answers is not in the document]
tomasatdatabricks: Actually I was wrong. Scala side has the same default. I got confused with the default params in the constructor. Good catch. I am moving it to the _from_java method.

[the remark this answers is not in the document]
tomasatdatabricks: Ok, I got doubly confused. I was right the first time -  DeepImageFeaturizer  in scala does *NOT HAVE DEFAULT* for input/output column.

HasInputCol and HasOutputCol traits have default value. Python wrapper extends these traits , however scala DeepImageFeaturizer does not inherit either because these are not accessible in scala. The end result is incompatible defaults between scala and python.

I think the only reasonable fix is to not too extend HasInput/HasOutput in python so that python and scala are consistent.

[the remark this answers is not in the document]
sueann: We actually copied over the python versions of `HasInputCol` and `HasOutputCol` from Spark to sparkdl, so we could do the same for Scala to ensure consistency.

python/sparkdl/transformers/named_image.py
```diff
@@ -140,17 +141,31 @@ def __call__(self, value):
...
     to the image column in DataFrame. The output is a MLlib Vector so that DeepImageFeaturizer
     can be used in a MLlib Pipeline.
     The input image column should be ImageSchema.
     """
-
+    # NOTE: We are not inheriting from HasInput/HasOutput to mirror the scala side. It also helps
+    # us to avoid issue with serialization/deserialization - default values based on uid do not
+    # always get set to correct value on the python side after deserialization. Default values do
+    # not get reset to the jvm side value unless they param value is not set.
+    # See pyspark.ml.wrapper.JavaParams._transfer_params_from_java
```
sueann: btw, is this really broken in Spark (maybe they have some workarounds)? wouldn't the output column param be very broken for all the transformers? or do some of them go through different code paths that avoid this error? ???

sueann: i think i do remember seeing some of the persistent code that reads in each param specifically from java to python so those probably avoid the issue. mostly trying to understand if there are any actual Transformers that have persistent broken currently in MLlib.

tomasatdatabricks: I think most of them do. It only matters if someone queries default params directly though 
 so it's probably not a top priority. Here's the repro with bucketizer:
```
from pyspark.ml.feature import Bucketizer
a = Bucketizer()
a.save("bucketizer0")
b = load("bucketizer0")
 a._defaultParamMap[a.outputCol]
u'Bucketizer_4b489ecc5cae3e86faf6__output'
b._defaultParamMap[b.outputCol]
u'Bucketizer_42388aff94842ab55f2f__output'
```

tomasatdatabricks: To make things worse:
```
 a.isSet(a.outputCol)
False
b.isSet(b.outputCol)
True
```

tomasatdatabricks: I've created a Jira for the issue https://issues.apache.org/jira/browse/SPARK-23244

sueann: Awesome, thanks!

PR #96, opened 2018-01-29
Bugfix: Properly expose KerasTransformer via 'from sparkdl import KerasTransformer'

smurching:
Fixes a bug where `KerasTransformer` wasn't imported  in our root __init__.py and so not accessible via `from sparkdl import KerasTransformer`

Tested by building the spark-deep-learning assembly JAR and running a KerasTransformer test against the JAR, using `from sparkdl import KerasTransformer`


sueann:
Thanks for fixing it! will merge.

PR #99, opened 2018-02-03
Fix bug in conversions from row image to/from BufferedImage

smurching:
Fix bug in conversions to/from BufferedImage in which we copied raw byte data to BufferedImage rasters using the wrong channel ordering. The fix in this PR is to use `BufferedImage.setRGB`, `BufferedImage.getRGB` APIs instead of accessing image raster data directly for three and four-channel images. 

Also enhanced an existing unit test to verify that we correctly convert from row image to BufferedImage for one and four-channel images.


smurching:
@sueann I think this is ready for review now, thanks in advance!

smurching:
Tested manually by resizing & displaying some images, the fix seems to work for me :)

thunterdb:
Thanks for the patch! Will this need fixes in Spark too?

On Fri, Feb 2, 2018 at 4:06 PM, Siddharth Murching <notifications@github.com

smurching:
Good question, we shouldn't need any fixes in Spark: this patch modifies Scala image-conversion utilities in DLP, which as of now are not part of Spark. It's a little confusing since DLP does contain Python image-conversion utils copied from Spark (in `python/sparkdl/image/image.py`) - these utils should probably be removed from DLP once Spark 2.3 is out.

PR #108, opened 2018-03-27
Add support for FloatType, IntegerType and LongType inputs to TFTransformer

smurching:
## What
Currently, TFTransformer accepts only DoubleType input columns, casting Double input into the input type expected by the TF graph. This casting implicitly constrains TFTransformer to support only those TensorFlow input types that can be casted-to from double.

This PR removes the cast-from-double operation in TFTransformer, thereby adding support for FloatType, IntegerType and LongType input columns.

## Why
The changes in this PR will facilitate adding support for BinaryType input columns in a follow-up

## Summary of Changes
* .gitignore: Add files generated by DeepLearningFeaturizer tests and codecov to .gitignore
* tf_tensor.py: Removes extra cast ops added to TF graphs in TFTransformer (**this is a breaking change**; before, users could feed Doubles into tf.float32, tf.int64, tf.int32 tensors). Instead, determines the type of each input tensor (in `_get_placeholder_types`) and replaces it with a placeholder of the same type using the TF `optimize_for_inference` API.
* keras_transformer_test.py: Changed to pass float32 input to Keras graphs (which have float32 input tensors) instead of float64 input.
* tf_transformer_test.py: Modified to test TFTransformer with float, int, and long input columns.

### DIFF
--- a/python/sparkdl/transformers/tf_tensor.py
+++ b/python/sparkdl/transformers/tf_tensor.py
@@ -20,7 +20,7 @@ from tensorflow.python.tools import optimize_for_inference_lib as infr_opt
 import tensorframes as tfs
 
 from pyspark.ml import Transformer
-from pyspark.sql.types import BooleanType
+from pyspark.sql.types import DoubleType
 
 from sparkdl.graph.builder import GraphFunction
 import sparkdl.graph.utils as tfx
--- a/python/sparkdl/transformers/tf_tensor.py
+++ b/python/sparkdl/transformers/tf_tensor.py
@@ -88,6 +88,13 @@ class TFTransformer(Transformer, HasTFInputGraph, HasTFHParams, HasInputMapping,
                                                placeholder_types)
 
     def _transform(self, dataset):
+        if len([field for field in dataset.schema if field.dataType == DoubleType()]) > 0:
+            logger.warn("Detected DoubleType columns in dataframe passed to transform(). In "
+                        "Deep Learning Pipelines 1.0 and above, DoubleType columns can only be "
+                        "fed to input tensors of type tf.float64. To feed dataframe data to "
+                        "tensors of other types (e.g. tf.float32, tf.int32, tf.int64), use the "
+                        "corresponding Spark SQL data types (FloatType, IntegerType, LongType).")
+
         graph_def = self._optimize_for_inference()
         input_mapping = self.getInputMapping()
         output_mapping = self.getOutputMapping()
--- a/python/sparkdl/transformers/tf_tensor.py
+++ b/python/sparkdl/transformers/tf_tensor.py
@@ -20,6 +20,7 @@ from tensorflow.python.tools import optimize_for_inference_lib as infr_opt
 import tensorframes as tfs
 
 from pyspark.ml import Transformer
+from pyspark.sql.types import BooleanType
 
 from sparkdl.graph.builder import GraphFunction
 import sparkdl.graph.utils as tfx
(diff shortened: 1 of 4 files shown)


python/sparkdl/transformers/tf_tensor.py
```diff
@@ -66,65 +63,29 @@ def setParams(self, tfInputGraph=None, inputMapping=None, outputMapping=None, tf
...
 
-    def _getSparkDlOpName(self, tensor_name):
-        """
-        Given a tensor name, returns the name of the op generating the tensor, prefixed with
-        a special scope indicating that the op has been added by Sparkdl.
-        """
-        op_name = tfx.op_name(tensor_name)
-        return tfx.add_scope_to_name(scope=self.SPARKDL_OP_SCOPE, name=op_name)
-
-    def _addCastOps(self, user_graph_def):
```
smurching: Remove logic for injecting casting ops before input tensors to convert from float64 -> expected data type

python/sparkdl/transformers/tf_tensor.py
```diff
@@ -66,65 +63,29 @@ def setParams(self, tfInputGraph=None, inputMapping=None, outputMapping=None, tf
...
         # Strip away graph nodes not used in computing the tensors with the specified output names
-        input_names = [self._getSparkDlOpName(tnsr_name) for _, tnsr_name in self.getInputMapping()]
+        input_names = [tfx.op_name(tnsr_name) for _, tnsr_name in self.getInputMapping()]
         output_names = [tfx.op_name(tnsr_name) for tnsr_name, _ in self.getOutputMapping()]
-        opt_gdef = infr_opt.optimize_for_inference(graphdef_with_casts,
-                                                   input_names,
-                                                   output_names,
-                                                   tf.float64.as_datatype_enum)
-        return opt_gdef
+        return infr_opt.optimize_for_inference(graph_def,
```
smurching: Replace each input tensor with a placeholder of the same type, instead of a `tf.float64` placeholder. See `optimize_for_inference` method signature: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/tools/optimize_for_inference_lib.py#L89

python/sparkdl/transformers/tf_tensor.py
```diff
@@ -137,12 +98,9 @@ def _transform(self, dataset):
             out_tnsr_op_names = [tfx.op_name(tnsr_name) for tnsr_name, _ in output_mapping]
             # Load graph
             tf.import_graph_def(graph_def=graph_def, name='', return_elements=out_tnsr_op_names)
-
             # Feed dict maps from placeholder name to DF column name
-            feed_dict = {self._getSparkDlOpName(
-                tnsr_name): col_name for col_name, tnsr_name in input_mapping}
+            feed_dict = {tfx.op_name(tnsr_name): col_name for col_name, tnsr_name in input_mapping}
```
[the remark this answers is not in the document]
smurching: Thanks for the review! Agree the `op_name` function is kinda confusing - in this PR we are just using it to strip the ":" part of the tensor name. Having a utility for getting the name of an op seems useful e.g. in case TensorFlow changes their op/tensor naming scheme, but I'm also open to not using it.

I think it'd be nice to:

* Simplify the `op_name` function so that it accepts just a tensor name instead of one of {tensor, tensor name, op name} - AFAICT this is possible without losing any existing functionality, but if not we could break `op_name` up into multiple utils.
* Rename usages of `tfx` -> `graph_utils` for clarity

For now maybe we can use `tfx.op_name` & eliminate all usages of `op_name` in a future PR?

smurching:
Chatted offline with @sueann, will optimistically merge this for now (with the only potential change being an error message update in a follow-up PR) - thanks @sueann @tomasatdatabricks!

smurching:
Chatted offline with @sueann - added a warning about the behavior change. Will merge now, thanks @sueann @tomasatdatabricks!

PR #109, opened 2018-03-28
Fix KerasImageFileEstimator for model tuning

yogeshg:
* add test for KerasImageFileEstimator used with CrossValidator
* fix pickling issue
* use new `fitMultiple` api and ensure thread safety
* fix tests to reflect new api
* bugfix setDefault
* bugfix HasOutputMode
* remove `_validateParams`
* avoid testing KIFT functionality in KIFEst tests

### DIFF
--- a/python/sparkdl/estimators/keras_image_file_estimator.py
+++ b/python/sparkdl/estimators/keras_image_file_estimator.py
@@ -17,15 +17,14 @@
 # pylint: disable=protected-access
 from __future__ import absolute_import, division, print_function
 
-import logging
 import threading
 
 import numpy as np
+from typing import Dict
 
+import pyspark
 from pyspark.ml import Estimator
 import pyspark.ml.linalg as spla
-from pyspark.ml.param import Param, Params, TypeConverters
-
 from sparkdl.image.imageIO import imageStructToArray
 from sparkdl.param import (
     keyword_only, CanLoadImage, HasKerasModel, HasKerasOptimizer, HasKerasLoss, HasOutputMode,
@@ -35,8 +34,6 @@ import sparkdl.utils.jvmapi as JVMAPI
 import sparkdl.utils.keras_model as kmutil
 
 __all__ = ['KerasImageFileEstimator']
-
-logger = logging.getLogger('sparkdl')
 
 
 class KerasImageFileEstimator(Estimator, HasInputCol, HasInputImageNodeName,
@@ -133,22 +130,16 @@ class KerasImageFileEstimator(Estimator, HasInputCol, HasInputImageNodeName,
         kwargs = self._input_kwargs
         return self._set(**kwargs)
 
-    def _validateParams(self):
+    def _validateParams(self, paramMap):
+        # type: (Dict[pyspark.ml.param.Param, object]) -> bool
         """
         Check Param values so we can throw errors on the driver, rather than workers.
         :return: True if parameters are valid
         """
-        if not self.isDefined(self.inputCol):
+        if not self.isDefined(self.inputCol) or self.inputCol in paramMap:
             raise ValueError("Input column must be defined")
-        if not self.isDefined(self.outputCol):
+        if not self.isDefined(self.outputCol) or self.outputCol in paramMap:
             raise ValueError("Output column must be defined")
-        return True
-
-    def _validateFitParams(self, params):
-        """ Check if an input parameter set is valid """
-        if isinstance(params, (list, tuple, dict)):
-            assert self.getInputCol() not in params, \
-                "params {} cannot contain input column name {}".format(params, self.getInputCol())
         return True
 
     def _getNumpyFeaturesAndLabels(self, dataset):
@@ -208,10 +199,11 @@ class KerasImageFileEstimator(Estimator, HasInputCol, HasInputImageNodeName,
         :param paramMaps: list of ParamMaps matching the maps in `kerasModelsRDD`
         :return: list of MLlib models
         """
-        # we can just use the ordering of rdd because they are in the same order as params map
-        for i, (param_map, model_bytes) in enumerate(kerasModelsBytesRDD.collect()):
-            model_filename = kmutil.bytes_to_h5file(model_bytes)
-            yield i, self._copyValues(KerasImageFileTransformer(modelFile=model_filename))
+        lock = threading.Lock()
+        for (i, model_bytes) in kerasModelsBytesRDD.collect():
+            with lock:
+                model_filename = kmutil.bytes_to_h5file(model_bytes)
+                yield i, self._copyValues(KerasImageFileTransformer(modelFile=model_filename))
 
     def fitMultiple(self, dataset, paramMaps):
         """
@@ -227,10 +219,15 @@ class KerasImageFileEstimator(Estimator, HasInputCol, HasInputImageNodeName,
                      existence of a sufficiently large (and writable) file system, users are
                      advised to not train too many models in a single Spark job.
         """
+        [self._validateParams(pm) for pm in paramMaps]
+
+        def _name_value_map(paramMap):
+            """takes a dictionary {param -> value} and returns a map of {param.name -> value}"""
+            return {param.name: val for param, val in paramMap.items()}
 
 
         sc = JVMAPI._curr_sc()
-        paramNameMaps = list(map(lambda x: {param.name: val for param, val in x.items()}, paramMaps))
+        paramNameMaps = list(enumerate(map(_name_value_map, paramMaps)))
         paramNameMapsRDD = sc.parallelize(paramNameMaps, numSlices=len(paramNameMaps))
 
         # Extract image URI from provided dataset and create features as numpy arrays
@@ -243,9 +240,8 @@ class KerasImageFileEstimator(Estimator, HasInputCol, HasInputImageNodeName,
         modelBytesBc = sc.broadcast(modelBytes)
 
         # Obtain params for this estimator instance
-        baseParamMap = self.extractParamMap()
-        baseParamDict = {param.name: val for param, val in baseParamMap.items()}
-        baseParamDictBc = sc.broadcast(baseParamDict)
+        baseParams = _name_value_map(self.extractParamMap())
+        baseParamsBc = sc.broadcast(baseParams)
 
         def _local_fit(override_param_map):
             """
@@ -256,7 +252,7 @@ class KerasImageFileEstimator(Estimator, HasInputCol, HasInputImageNodeName,
             :return: serialized Keras HDF5 file bytes
             """
             # Update params
-            params = baseParamDictBc.value
+            params = baseParamsBc.value
             params.update(override_param_map)
 
             # Create Keras model
@@ -271,7 +267,7 @@ class KerasImageFileEstimator(Estimator, HasInputCol, HasInputImageNodeName,
 
             return kmutil.model_to_bytes(model)
 
-        kerasModelBytesRDD = paramNameMapsRDD.map(lambda paramMap: (paramMap, _local_fit(paramMap)))
+        kerasModelBytesRDD = paramNameMapsRDD.map(lambda (i, paramMap): (i, _local_fit(paramMap)))
         return self._collectModels(kerasModelBytesRDD)
 
     def _loadModelAsBytes(self):
@@ -285,5 +281,5 @@ class KerasImageFileEstimator(Estimator, HasInputCol, HasInputImageNodeName,
         return fileContent
 
     def _fit(self, dataset):
-        tuples = self.fitMultiple(dataset, [self.extractParamMap()])
+        tuples = self.fitMultiple(dataset, [{}])
         return dict(tuples)[0]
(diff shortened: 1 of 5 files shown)


yogeshg:
* add test for KerasImageFileEstimator used with CrossValidator
the test should fail.

python/tests/estimators/test_keras_estimators.py
```diff
@@ -79,40 +89,63 @@ def _get_estimator(self, model, optimizer='adam', loss='categorical_crossentropy
...
+            .addGrid(estimator.kerasFitParams, [{"batch_size": 32}, {"batch_size": 64}])
+            .build()
+        )
+
+        bc = BinaryClassificationEvaluator(rawPredictionCol=self.output_col, labelCol=self.label_col)
+        cv = CrossValidator(estimator=estimator, estimatorParamMaps=paramGrid, evaluator=bc, numFolds=2)
+
+        transformer = cv.fit(image_uri_df)
+        self.assertIsInstance(transformer.bestModel, KerasImageFileTransformer, "best model should be KIFT")
+        output_df = transformer.transform(image_uri_df)
```
smurching: Here I think we can rely on the KIFT tests to confirm that `transform` produces the right result given a well-configured transformer. Maybe instead we should validate the params of the KIFT are set correctly (e.g. that it has the correct input & output columns?)

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -241,23 +208,30 @@ def _collectModels(self, kerasModelsBytesRDD):
         :param paramMaps: list of ParamMaps matching the maps in `kerasModelsRDD`
         :return: list of MLlib models
         """
-        transformers = []
-        for (param_map, model_bytes) in kerasModelsBytesRDD.collect():
+        # we can just use the ordering of rdd because they are in the same order as params map
+        for i, (param_map, model_bytes) in enumerate(kerasModelsBytesRDD.collect()):
```
smurching: Do you know if this assumption always holds? IIRC Spark doesn't tend to reorder results unless necessary, but it seems like it'd be safer to create your make your original RDD an RDD of `(paramMap, paramMapIndex)` and then pass the index through the map calls.

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -241,23 +208,30 @@ def _collectModels(self, kerasModelsBytesRDD):
...
             model_filename = kmutil.bytes_to_h5file(model_bytes)
-            transformers.append({
-                'paramMap': param_map,
-                'transformer': KerasImageFileTransformer(modelFile=model_filename)})
-
-        return transformers
+            yield i, self._copyValues(KerasImageFileTransformer(modelFile=model_filename))
```
smurching: Just one concern here, it looks like PySpark expects `fitMultiple` to return a threadsafe iterable ([see fitMultiple doc here](https://github.com/apache/spark/blob/30fcdc0/python/pyspark/ml/base.py#L90)) & by default achieves this through its own (internal) threadsafe iterator implementation, [see here](https://github.com/apache/spark/blob/30fcdc0/python/pyspark/ml/base.py#L30). I forget if PySpark prescribes a way developers should ensure thread safety, we can ask Joseph.

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -131,41 +133,6 @@ def setParams(self, inputCol=None, inputImageNodeName=None, outputCol=None,
...
-
-        :param dataset: input dataset, which is an instance of :py:class:`pyspark.sql.DataFrame`.
-                        The column `inputCol` should be of type `sparkdl.image.imageIO.imgSchema`.
-        :param params: An optional param map that overrides embedded params. If a list/tuple of
-                       param maps is given, this calls fit on each param map and returns a list of
-                       models.
-        :return: fitted model(s). If params includes a list of param maps, the order of these
-                  models matches the order of the param maps.
-        """
-        self._validateParams()
```
smurching: Currently this is unused - let's call this in `fitMultiple` to validate our params before running any Spark jobs?

Side note: it's a little weird that `_validateParams` only checks the input & output column parameters, it'd be nice to also validate that the other params are specified.

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -131,41 +133,6 @@ def setParams(self, inputCol=None, inputImageNodeName=None, outputCol=None,
...
-            if len(params) == 0:
-                paramMaps = [dict()]
-            else:
-                self._validateFitParams(params)
-                paramMaps = params
-        elif isinstance(params, dict):
-            self._validateFitParams(params)
-            paramMaps = [params]
-        else:
-            raise ValueError("Params must be either a param map or a list/tuple of param maps, "
```
smurching: Just to confirm: looks like in this PR we make a few breaking API changes:
1) We now require `KerasFitParams` to be a single param map dictionary & no longer support passing a list/tuple of param maps. 
2) `fit()` no longer accepts a `params` argument. 

For 1), I'm not sure if we need to support setting `KerasFitParams` to a list/tuple of param maps. IMO the question is whether using `KerasImageFileEstimator` without cross-validation is a legitimate use case (e.g. for hyperparameter tuning without a validation set). If it is legitimate, we should support passing a list/tuple (preferably just one of the two to simplify things for us, since we have an opportunity to break APIs with the DLP 1.0 release). If not, we can break the old behavior & should probably document that KIFE is intended to be used with CrossValidator. Curious to hear what @sueann & @tomasatdatabricks think here.

For 2) I'm fine with making this change - maybe for now we can keep `def fit(self, dataset, params=None)` and raise an exception describing the behavior change & how to deal with it if the user specifies `params`. We'll miss the edge case (i.e. not raise an exception) where users explicitly pass params=None, but hopefully almost nobody is doing that.

Let me know if I missed something!

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -131,41 +133,6 @@ def setParams(self, inputCol=None, inputImageNodeName=None, outputCol=None,
...
-        if params is None:
-            paramMaps = [dict()]
-        elif isinstance(params, (list, tuple)):
-            if len(params) == 0:
-                paramMaps = [dict()]
-            else:
-                self._validateFitParams(params)
-                paramMaps = params
-        elif isinstance(params, dict):
-            self._validateFitParams(params)
```
smurching: Remove `_validateFitParams` since it's currently unused? Also, do you know why we needed to check (in `_validateFitParams`) that the value of the input col param is not the name of one of the user-specified params? It seems like an unnecessary check to me.

python/tests/estimators/test_keras_estimators.py
yogeshg: we should validate the params of the KIFT are set correctly (e.g. that it has the correct input & output columns?)

python/sparkdl/estimators/keras_image_file_estimator.py
yogeshg: ah! perhaps something like the following?
```python
# somewhere
map2Index = dict(enumerate(paramMaps))
# in the for loop
i = map2Index[param_map]
```

I agree it is definitely safer and that the ordering guarantee is not explicitly documented. But I'd also be scared if spark messes this up:
```
anotherRDD = anRDD.map(someFunction)
```
link to Spark Scaladoc > Dataset > Map :
https://spark.apache.org/docs/latest/api/scala/index.html#org.apache.spark.sql.Dataset@map[U](func:org.apache.spark.api.java.function.MapFunction[T,U],encoder:org.apache.spark.sql.Encoder[U]):org.apache.spark.sql.Dataset[U]


@smurching, What are your thoughts?

python/sparkdl/estimators/keras_image_file_estimator.py
yogeshg: 1. [KerasFitParams] is a dict that contains hyperparameters that Keras needs (these are independent of MLlib `Param`s)
2. [estimator:fit] calls `_fit` (abstract implemented) or `fitMultiple` (if not overridden, [inherited]) depending on what the type of `params` is.
>         :param params: an optional param map that overrides embedded params. If a list/tuple of
>                       param maps is given, this calls fit on each param map and returns a list of
>                       models.



[KerasFitParams]:https://github.com/databricks/spark-deep-learning/blob/aeff9c9e2c24ed09c093414a7b51ec9bd42f1e86/python/sparkdl/param/shared_params.py#L150-L151
[estimator:fit]:https://github.com/apache/spark/blob/master/python/pyspark/ml/base.py#L110-L135
[inherited]:https://github.com/apache/spark/blob/30fcdc0/python/pyspark/ml/base.py#L90-L108

smurching: Oops you're totally right, didn't realize `fit` had a params argument :P - thanks for clarifying & sorry for the confusion!

python/sparkdl/estimators/keras_image_file_estimator.py
yogeshg: the logic `_validateFitParams` seems not useful and buggy, I've removed that.

@sueann @jkbradley @tomasatdatabricks  please note this change :)

python/sparkdl/estimators/keras_image_file_estimator.py
yogeshg: The original intention is to just check if `inputCol` and `outputCol` are set. Other `Param`s do not seem necessary to be checked, eg. `HasKerasModel` and other `HasKeras*` parameters seem to be used differently than the usual `Param`s.

Should we make a jira issue to check these, since that requires considerable work and not captured here? Please advise! Thanks! :)

yogeshg:
@smurching thanks for the review and help offline!!
I added the requested changes, and made a few comments to your questions.

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -241,23 +199,36 @@ def _collectModels(self, kerasModelsBytesRDD):
...
         """
-        transformers = []
-        for (param_map, model_bytes) in kerasModelsBytesRDD.collect():
-            model_filename = kmutil.bytes_to_h5file(model_bytes)
-            transformers.append({
-                'paramMap': param_map,
-                'transformer': KerasImageFileTransformer(modelFile=model_filename)})
+        lock = threading.Lock()
+        for (i, model_bytes) in kerasModelsBytesRDD.collect():
+            with lock:
```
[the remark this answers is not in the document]
yogeshg: again a good point, i need to think about this more thoroughly!

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -131,59 +130,18 @@ def setParams(self, inputCol=None, inputImageNodeName=None, outputCol=None,
...
-    def _validateParams(self):
+    def _validateParams(self, paramMap):
+        # type: (Dict[pyspark.ml.param.Param, object]) -> bool
         """
         Check Param values so we can throw errors on the driver, rather than workers.
         :return: True if parameters are valid
         """
-        if not self.isDefined(self.inputCol):
+        if not self.isDefined(self.inputCol) or self.inputCol in paramMap:
             raise ValueError("Input column must be defined")
```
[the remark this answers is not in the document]
yogeshg: the intent is that either `inputCol` is defined while initial constructor or `setParam` calls, or sent during parameter tuning in the paramMap. I'll put correct parens to clarify the intent `not (A or B in C)`.

For error message, how about:
"Input column must be defined or fine-tuned"
(It does still sound a bit curt)

yogeshg:
- [ ] keep the newlines between import groups ln:32
- [ ] thread safety
- [ ] raise issue for adding validations for the other params
- [ ] put parens `not (A or B in C)`
- [ ] "Input column must be defined or fine-tuned" or better

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -131,59 +130,18 @@ def setParams(self, inputCol=None, inputImageNodeName=None, outputCol=None,
...
-        if not self.isDefined(self.outputCol):
+        if not self.isDefined(self.outputCol) or self.outputCol in paramMap:
```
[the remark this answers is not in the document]
yogeshg: answered below

yogeshg:
- [ ] ensure parameters copied correctly

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -297,10 +303,12 @@ def _local_fit(override_param_map):
             _fit_params = params['kerasFitParams']
             model.fit(x=features, y=labels, **_fit_params)
 
-            return kmutil.model_to_bytes(model)
+            return index, override_param_map, kmutil.model_to_bytes(model)
+
+        kerasModelBytesRDD = paramNameMapsRDD.map(_local_fit)
+        models = self._collectModels(kerasModelBytesRDD)
```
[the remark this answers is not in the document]
smurching: ^Discussed offline, the code is good the way it is -- there's extra logic in `_collectModels` (i.e. it's not just wrapping `kerasModelBytesRDD.collect()` in a generator) that I overlooked.

PR #116, opened 2018-04-11
add better parameter validation for KIFEst

yogeshg:
* check if un-tunable parameters need to be defined
* check if tunable parameters need to be defined or specified in param map (this check against all those given to fitMultiple)
* add test to see if method works correctly

### DIFF
--- a/python/sparkdl/estimators/keras_image_file_estimator.py
+++ b/python/sparkdl/estimators/keras_image_file_estimator.py
@@ -167,20 +167,22 @@ class KerasImageFileEstimator(Estimator, HasInputCol, HasOutputCol, HasLabelCol,
         :param paramMap: Dict[pyspark.ml.param.Param, object]
         :return: True if parameters are valid
         """
-        input_params = ['imageLoader', 'inputCol', 'labelCol']
-        model_params_untunable = ['modelFile']
-        model_params = ['kerasOptimizer', 'kerasLoss', 'kerasFitParams']
-        output_params = ['outputCol', 'outputMode']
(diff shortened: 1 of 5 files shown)


python/tests/estimators/test_keras_estimators.py
```diff
@@ -105,7 +103,27 @@ def setUp(self):
     def tearDown(self):
         shutil.rmtree(self.temp_dir, ignore_errors=True)
 
+    def test_validate_params(self):
+        """Test that `KerasImageFileEstimator._validateParams` method works as expected"""
```
sueann: we probably could be more thorough in making sure all the intersect/diff logic works but I think it's good enough.

python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -174,12 +167,23 @@ def _validateParams(self, paramMap):
         :param paramMap: Dict[pyspark.ml.param.Param, object]
         :return: True if parameters are valid
         """
-        if not self.isDefined(self.inputCol):
-            raise ValueError("Input column must be defined")
-        if not self.isDefined(self.outputCol):
-            raise ValueError("Output column must be defined")
-        if self.inputCol in paramMap:
-            raise ValueError("Input column can not be fine tuned")
+        model_params = [self.kerasOptimizer, self.kerasLoss, self.kerasFitParams]
```
sueann: merge `model_params` and `output_params` into `tunable_params` for clarity

yogeshg:
Offline review with @smurching , add error for untunable param.

raised raised https://github.com/databricks/spark-deep-learning/pull/117/ in case this becomes a blocker and we are good with merging as is.

PR #119, opened 2018-04-17
Replaced copy-pasted image schema and all references  with the original

tomasatdatabricks:
We do not support Spark<2.3 anymore so there is no need for the copy-pasted ImageSchema file.


smurching:
Let's also remove `sparkdl.readImages` & related methods ('imageSchema', 'imageType', 'readImages') from python/sparkdl/__init__.py? Otherwise LGTM

PR #120, opened 2018-04-20
broadcast only tunable parameters

yogeshg:
We want to broadcast only the parameters that are used in tuning. Other parameters like "imageLoader" may not be required, and throw error if they are not pickle-able.

To do this, we use the `tunable_params` that we already used to validate params, it has been sent to the constructor, and is also used at the time of broadcast.

To test this, we ran the local tests, could not run the local-cluster test since it does not support pyspark (thanks for help, Lu Wang). To test on cluster, we used Databricks clusters.

### DIFF
--- a/python/sparkdl/estimators/keras_image_file_estimator.py
+++ b/python/sparkdl/estimators/keras_image_file_estimator.py
@@ -148,6 +148,8 @@ class KerasImageFileEstimator(Estimator, HasInputCol, HasOutputCol, HasLabelCol,
         super(KerasImageFileEstimator, self).__init__()
         kwargs = self._input_kwargs
         self.setParams(**kwargs)
+        self._tunable_params = [self.kerasOptimizer, self.kerasLoss, self.kerasFitParams,
+                          self.outputCol, self.outputMode]  # model params and output params
 
     @keyword_only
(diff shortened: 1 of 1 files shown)


python/sparkdl/estimators/keras_image_file_estimator.py
```diff
@@ -285,7 +284,8 @@ def _name_value_map(paramMap):
         modelBytesBc = sc.broadcast(modelBytes)
 
         # Obtain params for this estimator instance
-        baseParams = _name_value_map(self.extractParamMap())
```
smurching: nit/suggestion: maybe for consistency `_name_value_map` can filter out untunable params & we can continue to use `_name_value_map` here? thinking it'd be nice since we use `_name_value_map` above to process `paramMaps` and it would be good to handle both `paramMaps` and `self.extractParamMap()` consistently.

for clarity (since `_name_value_map` would now be doing more than just getting a name -> value map) we could rename `_name_value_map` -> `_get_tunable_name_value_map` or something

yogeshg: @smurching, thanks for catching this!
@sueann, Yes! that's what we check in '_validateParams'.

sueann:
cool. lgtm as well. feel free to merge once the tests pass @smurching

PR #137, opened 2018-06-15
Fix type error in TFTransformer example in README

statX:
Fixes the following Exception that occurs when running ```odf = transformer.transform(sdf)``` with Apache Spark 2.3.0, Scala 2.11 from Databricks

```java.lang.Exception: The type of node &apos;input_tensor&apos; (ScalarDoubleType) is not compatible with the data type of the column (ScalarFloatType)```


sarchak:
Hitting the same issue any plans on merging this?

ronitshaw:
Facing the same issue!

PR #149, opened 2018-08-14
[WIP] drop Spark <2.3 support and build against Keras 2.2.2 and TF 1.10.0

mengxr:
Still a WIP.

* Drop support for Spark <2.3 and hence Scala 2.10.
* Update Keras version to 2.2.2 and TF version to 1.10.0.
* TODO: Update README

We will add Spark 2.4 support after RC is out.

PR #151, opened 2018-08-16
Support and build against Keras 2.2.2 and TF 1.10.0

ludatabricks:
* bump spark version to `2.3.1`
* bump tensorframes version to `0.4.0`
* bump `keras==2.2.2` and `tensorflow==1.10.0` to fix travis issues
* TF_C_API_GRAPH_CONSTRUCTION added as a temp fix
* Drop support for Spark <`2.3` and hence Scala `2.10`
* add python3 friendly print
* add `pooling='avg'` in resnet50 testing model beccause keras api changed
* test arrays almost equal with whatever precision 5 in `NamedImageTransformerBaseTestCase`, `test_bare_keras_module`, `keras_load_and_preproc`
* make keras model smaller in `test_simple_keras_udf`

This is a continued work from https://github.com/databricks/spark-deep-learning/pull/149.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ env:
     - SPARK_BUILD_URL="https://dist.apache.org/repos/dist/release/spark/spark-2.3.1/spark-2.3.1-bin-hadoop2.7.tgz"
     - SPARK_HOME=$HOME/.cache/spark-versions/$SPARK_BUILD
     - RUN_ONLY_LIGHT_TESTS=True
+    - TF_C_API_GRAPH_CONSTRUCTION=0
   matrix:
     - PYTHON_VERSION=3.6.2 TEST_SUITE=scala-tests
     - PYTHON_VERSION=3.6.2 TEST_SUITE=python-tests
@@ -50,6 +51,7 @@ before_install:
                -e PYSPARK_PYTHON
                -e SPARK_HOME
                -e RUN_ONLY_LIGHT_TESTS
+               -e TF_C_API_GRAPH_CONSTRUCTION
                -e CONDA_URL
                -d --name ubuntu-test -v $HOME ubuntu:16.04 tail -f /dev/null
   - docker ps
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,20 +18,20 @@ env:
     - SCALA_VERSION=2.11.8
     - SPARK_VERSION=2.3.1
     - SPARK_BUILD="spark-${SPARK_VERSION}-bin-hadoop2.7"
-    - SPARK_BUILD_URL="https://dist.apache.org/repos/dist/release/spark/spark-2.3.1/spark-2.3.1-bin-hadoop2.7.tgz"
+    - SPARK_BUILD_URL="https://dist.apache.org/repos/dist/release/spark/spark-${SPARK_VERSION}/spark-${SPARK_VERSION}-bin-hadoop2.7.tgz"
     - SPARK_HOME=$HOME/.cache/spark-versions/$SPARK_BUILD
     - RUN_ONLY_LIGHT_TESTS=True
     # TODO: This is a temp fix in order to pass tests.
     # We should update implementation to allow graph construction via C API.
     - TF_C_API_GRAPH_CONSTRUCTION=0
   matrix:
-    # - PYTHON_VERSION=3.6.2 TEST_SUITE=scala-tests
-    # - PYTHON_VERSION=3.6.2 TEST_SUITE=python-tests
-    # - PYTHON_VERSION=3.6.2 TEST_SUITE=pylint
-    # - PYTHON_VERSION=3.5.1 TEST_SUITE=python-tests
-    # - PYTHON_VERSION=3.5.1 TEST_SUITE=pylint
+    - PYTHON_VERSION=3.6.2 TEST_SUITE=scala-tests
+    - PYTHON_VERSION=3.6.2 TEST_SUITE=python-tests
+    - PYTHON_VERSION=3.6.2 TEST_SUITE=pylint
+    - PYTHON_VERSION=3.5.1 TEST_SUITE=python-tests
+    - PYTHON_VERSION=3.5.1 TEST_SUITE=pylint
     - PYTHON_VERSION=2.7.13 TEST_SUITE=python-tests
-    # - PYTHON_VERSION=2.7.13 TEST_SUITE=pylint
+    - PYTHON_VERSION=2.7.13 TEST_SUITE=pylint
 
 before_install:
   - ./bin/download_travis_dependencies.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ env:
     - SCALA_VERSION=2.11.8
     - SPARK_VERSION=2.3.1
     - SPARK_BUILD="spark-${SPARK_VERSION}-bin-hadoop2.7"
-    - SPARK_BUILD_URL="https://dist.apache.org/repos/dist/release/spark/spark-${SPARK_VERSION}/spark-${SPARK_VERSION}-bin-hadoop2.7.tgz"
+    - SPARK_BUILD_URL="https://dist.apache.org/repos/dist/release/spark/spark-2.3.1/spark-2.3.1-bin-hadoop2.7.tgz"
     - SPARK_HOME=$HOME/.cache/spark-versions/$SPARK_BUILD
     - RUN_ONLY_LIGHT_TESTS=True
   matrix:
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,13 +25,13 @@ env:
     # We should update implementation to allow graph construction via C API.
     - TF_C_API_GRAPH_CONSTRUCTION=0
   matrix:
-    - PYTHON_VERSION=3.6.2 TEST_SUITE=scala-tests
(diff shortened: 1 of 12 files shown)


ludatabricks:
retest

python/tests/graph/test_pieces.py
```diff
@@ -169,6 +169,6 @@ def test_pipeline(self):
                 # tfx.write_visualization_html(issn.graph,
                 # NamedTemporaryFile(prefix="gdef", suffix=".html").name)
 
-            self.assertTrue(np.all(preds_tgt == preds_ref))
+            np.testing.assert_array_almost_equal(preds_tgt, preds_ref, decimal=5)
```
yogeshg: can we use `featurizerCompareDigitsExact` setting centrally, if we want to control this?

python/tests/transformers/named_image_ResNet50_test.py
```diff
@@ -20,3 +20,4 @@ class NamedImageTransformerResNet50Test(NamedImageTransformerBaseTestCase):
 
     __test__ = True
     name = "ResNet50"
+    poolingMethod = 'avg'
```
yogeshg: can we comment why this is necessary in the description of this PR?

python/tests/udf/keras_sql_udf_test.py
```diff
@@ -60,10 +60,8 @@ def test_simple_keras_udf(self):
         with IsolatedSession(using_keras=True) as issn:
             model = Sequential()
             model.add(Flatten(input_shape=(640, 480, 3)))
-            model.add(Dense(units=64))
-            model.add(Activation('relu'))
-            model.add(Dense(units=10))
-            model.add(Activation('softmax'))
+            # model.add(Dense(64, activation='relu'))
+            model.add(Dense(16, activation='softmax'))
```
yogeshg: can we leave a comment here suggesting why we removed these lines?

.travis.yml
```diff
@@ -21,6 +21,9 @@ env:
     - SPARK_BUILD_URL="https://dist.apache.org/repos/dist/release/spark/spark-${SPARK_VERSION}/spark-${SPARK_VERSION}-bin-hadoop2.7.tgz"
     - SPARK_HOME=$HOME/.cache/spark-versions/$SPARK_BUILD
     - RUN_ONLY_LIGHT_TESTS=True
+    # TODO: This is a temp fix in order to pass tests.
+    # We should update implementation to allow graph construction via C API.
+    - TF_C_API_GRAPH_CONSTRUCTION=0
```
yogeshg: For more context, IIRC, i added this as a debug mechanism, i do not think this should make a condition for our testing, according to the commit, https://github.com/tensorflow/tensorflow/commit/0f60a7d10102048625e45c07a59c71c0c4375529 
> One known difference is improved static shape inference, meaning some shape errors will be surfaced during graph construction instead of at runtime.

Which means we might need to update our assumption that the shape will be checked only during run time. As noted in the comments above.

Also, i remember a vague suspicion that we may have ben using different versions of tensorflow in tensorframes and this library. can we confirm if both tensorframes and sparkdl use tf 1.10 ?

python/tests/transformers/named_image_test.py
```diff
@@ -69,8 +69,9 @@ class NamedImageTransformerBaseTestCase(SparkDLTestCase):
     name = None
     # Allow subclasses to force number of partitions - a hack to avoid OOM issues
     numPartitionsOverride = None
-    featurizerCompareDigitsExact = 6
+    featurizerCompareDigitsExact = 5
```
yogeshg: one easy way to make this global is by adding an `assertArrayAlmostEqual` method to `SparkDLTestCase` that calls `np.testing.test_arrays_almost_equal` and takes a default comparing precision, and using that instead of assertEquals. How do you feel about this?

python/tests/transformers/named_image_test.py
```diff
@@ -69,8 +69,9 @@ class NamedImageTransformerBaseTestCase(SparkDLTestCase):
...
     featurizerCompareDigitsCosine = 1
+    poolingMethod = None
```
yogeshg: Is it true that pooling is None in testing and `avg` in prod? if so, can we clarify why in the description?

.travis.yml
ludatabricks: We didn't release tensorframes with TF 1.10 yet. After the releasing of the new tensorframes we can retest this.

python/tests/transformers/named_image_test.py
ludatabricks: Good idea. It may needs to change all of the usage of `np.testing.test_arrays_almost_equal`. It might be better to change it on a different PR.

python/tests/transformers/named_image_test.py
ludatabricks: For other models, we can use the default poolingMethod which is None. Only issue is ResNet50. The new Keras version changed the model. In order to make it match as what we did before, we need to add the avg pooling in ResNet50 model.

python/requirements.txt
```diff
@@ -1,11 +1,11 @@
...
 coverage>=4.4.1
 h5py>=2.7.0
-keras==2.1.5 # NOTE: this package has only been tested with keras 2.1.5 and may not work with other releases
+keras==2.2.2 # NOTE: this package has only been tested with keras 2.2.2
 nose>=1.3.7  # for testing
 parameterized>=0.6.1 # for testing
 pillow>=4.1.1,<4.2
 pygments>=2.2.0
-tensorflow==1.6.0
+tensorflow==1.10.0 # NOTE: this package has only been tested with tensorflow 0.10.0
```
jkbradley: In note: 0.10.0 -> 1.10.0

python/tests/transformers/named_image_test.py
yogeshg: 1. I was looking at the test cases, can you point me to which one is it that fails?
If I understand correctly, we check the output of this model, as stored in `kerasPredict` against tensorflow, resized images and deepimagepredictor model. Shouldn't all of them be still the same? if not, which one and why?
2. also, should we simply add `pooling="avg"` in https://github.com/databricks/spark-deep-learning/blob/4a2ae917630dfc4aa7a9921d9810debb79991814/python/sparkdl/transformers/keras_applications.py#L228? to achieve the same effect as you are achieving without adding the testing harness to test against all poolings?

python/tests/graph/test_pieces.py
```diff
@@ -49,6 +49,8 @@
 
 
 class GraphPiecesTest(SparkDLTestCase):
+    
+    featurizerCompareDigitsExact = 5
```
yogeshg: I was thinking we could put this all the way in `SparkDLTestCase` and use it everywhere we do the array almost comparison. But if this is blocking and we want to put that off for later, fine by me.

yogeshg:
Thanks for the update, Lu! As a summary of my comments:
* [ ] can we put the testingPrecision in SparkDLTestCase ?
* [ ] which test case fails with the standard setting of resnet50? and follow up questions inline.
* [ ] do you think we need to make matching changes in `model` and `_testKerasModel` methods of resnet50?
* [ ] `assertArrayAlmostEqual` to be introduced in a follow up pr
* [ ] `tensorframes` release to follow or precede this release

python/tests/transformers/named_image_test.py
ludatabricks: The test `test_featurization_no_reshape` fails because kerasReshaped and dfFeatures has different shapes.

https://github.com/databricks/spark-deep-learning/blob/master/python/tests/transformers/named_image_test.py#L186

python/tests/udf/keras_sql_udf_test.py
ludatabricks: Made one comments to explain it.

python/tests/graph/test_pieces.py
ludatabricks: It might not be a good idea to put the featurizerCompareDigitsExact in SparkDLTestCase. Because of the different condition numbers, we might need different number for different model. Like for  InceptionV3 and Xception, we make it to 4. Since `np.testing.test_arrays_almost_equal` already have a default value 6. We may not want to define a different one, unless we define the new  `assertArrayAlmostEqual ` as you suggested.

python/tests/transformers/named_image_test.py
ludatabricks: About the second. My question is do we want to support other pooling method for Kreas model in the future?

python/tests/graph/test_pieces.py
yogeshg: Aha! did not know that. that is definitely more complicated.

python/sparkdl/transformers/keras_applications.py
```diff
@@ -188,7 +188,8 @@ def inputShape(self):
         return (299, 299)
 
     def _testKerasModel(self, include_top):
-        return xception.Xception(weights="imagenet", include_top=include_top)
+        return xception.Xception(weights="imagenet",
```
yogeshg: nit: we do not need the new line here, but that cool.

yogeshg:
LGTM pending tests. @jkbradley , LGTY?

yogeshg:
LGTM pending tests. @jkbradley , LGTY?

jkbradley:
LGTM too, thanks!

jkbradley:
I'll merge this now.  Thanks @ludatabricks and @yogeshg !

PR #156, opened 2018-09-06
Add horovod runner api

yogeshg:


python/tests/horovod/runner_test.py
```diff
@@ -0,0 +1,78 @@
...
+import contextlib
+import logging
+import sys
+
+from six import StringIO
+
+from sparkdl import HorovodRunner
+from ..tests import PythonUnitTestCase
+
+
```
yogeshg: These patches would be so much easier had we been testing only against python3. just saying :P

.travis.yml
```diff
@@ -75,19 +80,20 @@ install:
...
       pip install --user -r $HOME/spark-deep-learning/python/requirements.txt;"
 
-script:
   - docker cp $HOME/.cache ubuntu-test:$HOME/
   # build assembly
   - docker exec -t ubuntu-test bash -c "
       source $HOME/miniconda/bin/activate test-environment;
       cd $HOME/spark-deep-learning;
       ./dev/run.py assembly"
+
```
yogeshg: I moved this here since this folds the above step in travis ci, saves us a lot of time scrolling. i do not feel strongly about this tho. cc @MrBago

python/sparkdl/horovod/runner.py
```diff
@@ -0,0 +1,56 @@
...
+class HorovodRunner(object):    # pylint: disable=too-few-public-methods
+    """
+    HorovodRunner can be used to run distributed workloads on a cluster.
+    Note that the open source version only has an interface and runs the workload in the current
+    driver only. Switch to Databricks to automatically run the workload distributedly.
+    """
+
+
+    @keyword_only
+    def __init__(self, np, timeline_path, retries):
```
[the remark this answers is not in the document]
yogeshg: ah, i think i should add a default value to `timeline_path` and `retries` both. great catch!

.travis.yml
```diff
@@ -22,13 +22,18 @@ env:
     - SPARK_HOME=$HOME/.cache/spark-versions/$SPARK_BUILD
     - RUN_ONLY_LIGHT_TESTS=True
   matrix:
+    # Run python horovod tests and style first
+    - PYTHON_VERSION=3.6.2 TEST_SUITE="python-tests python/tests/horovod/"
+    - PYTHON_VERSION=3.6.2 TEST_SUITE=pylint
+    - PYTHON_VERSION=3.5.1 TEST_SUITE="python-tests python/tests/horovod/"
+    - PYTHON_VERSION=3.5.1 TEST_SUITE=pylint
+    - PYTHON_VERSION=2.7.13 TEST_SUITE="python-tests python/tests/horovod/"
```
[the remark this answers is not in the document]
yogeshg: We unfortunately have to, since this will also run on python2 supported clusters. I do agree things would be much easier and nicer and neater if we had only python3

python/tests/horovod/runner_test.py
yogeshg: cc @ludatabricks

PR #157, opened 2018-09-07
update travis for quick development on horovod runner

yogeshg:


jkbradley:
LGTM
Travis will not succeed since we don't have horovod_runner tests yet, but that's fine.
I'll merge this now.

PR #158, opened 2018-09-18
fix the assembly artifact name

mengxr:
The assembly name is currently based on the project dir name. We can define `name` in the sbt file to make it deterministic.

cc: @yogeshg


mengxr:
The change is in master.

PR #161, opened 2018-10-11
Remove persistence from DeepImageFeaturizer

jkbradley:
Until Apache Spark provides complete support for persistence in 3rd-party Python libraries, we need to remove this buggy implementation of persistence.  Users have exported Pipelines with DeepImageFeaturizer, expecting to be able to import them back but finding they cannot.


ludatabricks:
LGTM

PR #165, opened 2018-11-14
[ML-5589.2] rename HorovodRunnerBase to HorovodRunner

mengxr:
So the API docs are not showing `HorovodRunner` as an alias of `HorovodRunnerBase`.

screen shot 2018-11-13 at 4 53 15 pm

TODOs
- [ ] update doc site


jkbradley:
LGTM pending tests

PR #169, opened 2018-11-18
[ML-5678] Use `conda activate` in docker for best compatibility

mengxr:
Use "conda activate" for the best compatibility with conda. This PR also made the following changes:

1. Set /mnt/sparkdl as the WORKDIR in Docker.
2. Mount ~/.ivy2 to accelerate sbt build in docker.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,10 @@ env:
 services:
   - docker
 
+cache:
+  directories:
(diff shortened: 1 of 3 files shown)


jkbradley:
LGTM
I mainly checked the test logs vs other logs, plus some travis & docker docs for reference.

I'll merge this now.

PR #171, opened 2018-11-20
Add multiple outputs to the TFImageTransformer (second attempt)

thunterdb:
This PR adds support for outputting multiple elements from the low-level TFImageTransformer. With this PR, one can then take TensorFlow's object detectors and run them at scale with Spark.

This adds a new output mode ("sql"): under this mode, the output of running tensorflow is added as extra columns to the resulting dataframe and is not interpreted as images or vectors.

Includes test.


python/sparkdl/param/image_params.py
```diff
@@ -25,7 +25,7 @@
 from sparkdl.image.imageIO import _reverseChannels, imageArrayToStruct
 from sparkdl.param.converters import SparkDLTypeConverters
 
-OUTPUT_MODES = ["vector", "image"]
+OUTPUT_MODES = ["vector", "image", "sql"]
```
mengxr: Where are the user doc for those modes and the new "sql" mode?

python/sparkdl/transformers/tf_image.py
```diff
@@ -195,33 +221,55 @@ def _addReshapeLayers(self, tf_graph, dtype="uint8"):
...
+                output_mapping = [(_getOriginalTensorName(tnsr_name),
+                                   _getFinalOpName(tnsr_name))
+                                  for (tnsr_name, _) in self.getOutputMapping()]
+            else:
+                output_mapping = [(self._getOriginalOutputTensorName(),
+                                   self._getFinalOutputOpName())]
+
+            for (tnsr_name, out_name) in output_mapping:
+                output_node = g.get_tensor_by_name(tnsr_name)
+                _ = tf.reshape(output_node[0], shape=[-1], name=out_name)
```
mengxr: Any reason to assign the return value to a single `_`?

python/tests/transformers/tf_image_test.py
```diff
@@ -176,3 +176,37 @@ def test_prediction_vs_tensorflow_inceptionV3(self):
...
+                # keras expects array in RGB order, we get it from image schema in BGR =>
+                # need to flip
+                preprocessed = preprocess_input(imageIO._reverseChannels(resized_images))
+                model = InceptionV3(input_tensor=preprocessed, weights="imagenet")
+                graph = tfx.strip_and_freeze_until([model.output], g, sess, return_graph=True)
+
+        transformer = TFImageTransformer(channelOrder='BGR', inputCol="image", outputCol=output_col, graph=graph,
+                                         inputTensor=image_string, outputTensor=model.output,
+                                         outputMode="sql",
+                                         outputMapping={model.output.name: 'my_prediction'})
```
mengxr: `'my_prediction'` -> `output_col`

python/tests/transformers/tf_image_test.py
```diff
@@ -176,3 +176,37 @@ def test_prediction_vs_tensorflow_inceptionV3(self):
...
+        transformed_df = transformer.transform(image_df.limit(10))
```
mengxr: I saw a limit here but no limit with `tf_values`. We don't have more than 10 images. But it is better to remove this limit or use the same limit in TensorFlow.

PR #178, opened 2018-12-06
[ML-5813] Add flag to release script for publishing docs.

MrBago:
This PR makes building and publishing docs optional with the release script.


jkbradley:
LGTM
I tested it locally
Merging with master
Thanks!

PR #181, opened 2018-12-28
[ML-5968] Update publishing dependencies

MrBago:
Use spDist with manual upload to publish instead of spPublish because the upload step of spDist is flaky.


yogeshg:
LGTM

PR #183, opened 2019-01-24
[ML-6097] Update Horovod Runner to include return value and multi-gpu info.

MrBago:
Cherry-picked the commit that adds return value implemenation, docs & tests.
Added better description of `np` to include info about multi-gpu support.


mengxr:
LGTM

PR #198, opened 2019-07-10
Update API doc for single-node distributed training

mengxr:


python/sparkdl/horovod/runner_base.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, np):
             which maps to a GPU on a GPU cluster or a CPU core on a CPU cluster.
             Accepted values are:
 
-            - If -1, this will spawn a subprocess on the driver node to run the Horovod job locally.
+            - If <0, this will spawn -np subprocesses on the driver node to run Horovod locally.
```
hanyucui: @mengxr I don't think "-np subprocesses" is straight forward for a layperson, do you think it's oK for our customers? Do you plan to do the release as well?

hanyucui: Here's the PR I have: https://github.com/databricks/spark-deep-learning/pull/201/files

PR #200, opened 2019-07-16
[ML-7565] Update library dependencies

mengxr:
Update the following libraries for MLR5.5:

Tensorflow=1.13.1
horovod=1.16.4
cloudpickle=0.8.0

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -12,10 +12,10 @@ dependencies:
 - pandas=0.23.4
 - h5py=2.8.0
 - pillow=4.1.1
-- cloudpickle=0.5.2
-- tensorflow=1.12.0
+- cloudpickle=0.8.0
+- tensorflow=1.13.1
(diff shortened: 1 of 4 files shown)


environment.yml
```diff
@@ -12,20 +12,22 @@ dependencies:
...
-- paramiko=2.4.1
+- paramiko=2.4.2
 - wrapt=1.10.11
 # test
 - coverage=4.5.1
 - nose=1.3.7
 - parameterized=0.6.1
 - pylint=1.8.4
 - argcomplete=1.9.4
+- pytorch-cpu=1.1.0
```
Loquats: I think we need to add the pytorch channel to install these two packages. https://anaconda.org/pytorch

mengxr: I removed pytorch because no tests require pytorch and it will increase the build time.

PR #233, opened 2020-08-25
[ML-11407] Update sparkdl OSS repo to 2.0.0 and only keep the dummy HorovodRunner

WeichenXu123:
Update sparkdl OSS repo to 2.0.0 and only keep the dummy HorovodRunner


LICENSE
```diff
@@ -84,4 +84,3 @@ To apply the Apache License to your work, attach the following boilerplate notic
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
-
```
liangz1, line 87: This deletion seems not necessary.

sparkdl/horovod/__init__.py
```diff
@@ -13,3 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+
+# pylint: disable=wrong-import-position
+# pylint: disable=invalid-name
```
liangz1, line 17: Why do we need these given there is no import or name here?

sparkdl/__init__.py
```diff
@@ -12,3 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+
+# pylint: disable=wrong-import-position
+# pylint: disable=invalid-name
```
liangz1, line 17: Why do we need these given there is no import here? What is the invalid name?

setup.py
```diff
@@ -0,0 +1,45 @@
...
+        "Environment :: Console",
+        "License :: OSI Approved :: BSD License",
+        "Operating System :: Unix",
+        "Programming Language :: Python :: 3",
+        "Topic :: Scientific/Engineering",
+        "Topic :: Software Development",
+    ],
+    platforms=["Linux"],
+    license="BSD",
+    keywords="spark deep learning horovod model distribbuted training",
```
liangz1, line 40: typo: distribbuted

run-tests.sh
```diff
@@ -101,9 +61,8 @@ for noseOptions in $noseOptionsArr
 do
   echo "============= Running the tests in: $noseOptions ============="
   # The grep below is a horrible hack for spark 1.x: we manually remove some log lines to stay below the 4MB log limit on Travis.
```
liangz1, line 63: Do we still support spark 1.x?

run-tests.sh
```diff
@@ -101,9 +61,8 @@ for noseOptions in $noseOptionsArr
...
-  $PYSPARK_DRIVER_PYTHON \
+  python \
```
liangz1, line 64: Do we need to enforce `python3` here?

setup.py
```diff
@@ -0,0 +1,45 @@
...
+
+setuptools.setup(
+    name='sparkdl',
+    version=sparkdl.__version__,
+    packages=setuptools.find_packages(exclude=exclude_packages),
+    url="https://github.com/databricks/spark-deep-learning",
+    author="Weichen Xu",
+    author_email="weichen.xu@databricks.com",
+    description="Deep Learning Pipelines for Apache Spark",
+    long_description="",
```
liangz1, line 26: Do we want to mention the fact that this package is only for testing purposes?

sparkdl/horovod/__init__.py
WeichenXu123: It is minor one, we can leave it to reduce conflicts.

sparkdl/__init__.py
WeichenXu123: ditto

LICENSE
WeichenXu123: ditto

PR #235, opened 2020-09-02
[ML-11959]  Create docs

WeichenXu123:
Create docs by sphinx.
image


liangz1:
We may want to remove references to MLR 5.0 in the docs since it's outdated.

PR #245, opened 2022-03-17
Use github actions to replace travis

yxiong:


.github/workflows/test.yml
```diff
@@ -0,0 +1,21 @@
...
+      - name: Check out repository code
+        uses: actions/checkout@v2
+      - name: Set up python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install sphinx==3.2.1 pyspark==3.0.1 tensorflow-cpu==1.15.0 pytest
```
lu-wang-dl, line 16: Update pyspark to 3.2.1? Shell we update other packages too?

.github/workflows/test.yml
```diff
@@ -0,0 +1,21 @@
...
+jobs:
+  run-unit-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v2
+      - name: Set up python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
```
lu-wang-dl, line 12: Maybe use 3.9?

.github/workflows/test.yml
```diff
@@ -0,0 +1,21 @@
+name: Test
+on: [push]
+jobs:
+  run-unit-test:
+    runs-on: ubuntu-latest
```
lu-wang-dl, line 5: Do we want to ping the ubuntu to a specific version?

.github/workflows/test.yml
```diff
@@ -0,0 +1,21 @@
+name: Test
+on: [push]
```
lu-wang-dl, line 2: Also run the test for other conditions like `pull_request, workflow_dispatch`

yxiong: Done.

.github/workflows/test.yml
yxiong: Done.

.github/workflows/test.yml
yxiong: Sounds good, upgraded all packages to the latest.

.github/workflows/test.yml
yxiong: Sure, pinned to ubuntu-20.04

yxiong:
Thanks for the review, PTAL.

.github/workflows/test.yml
```diff
@@ -0,0 +1,21 @@
+name: Test
+on: [push, pull_request, workflow_dispatch]
```
wenfeiy-db, line 2: Just curious. I saw two checks (one said push and the other said pull request) in this PR have been run, is this because we check on both `push` and `pull_request`? Will we always run the same test twice on a new commit in a PR? Is this expected?

yxiong: This is a list suggested by @lu-wang-dl . Here is the definition of events: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows

I think `push` is enough, and if we run test on every commit, `pull_request` is not necessary. I'll update the list to `[push, workflow_dispatch]`. @lu-wang-dl what do you think?

lu-wang-dl: I think the difference is on a PR from a forked repo. push event may not able to trigger the run then. You can limit the runs to master like:
```
on:
  push:
    branches:
    - master
  pull_request:
    branches:
    - master
```

yxiong: I see. The reason it triggered twice is that I was too lazy to fork the repository. In that case I think we can just have both push and pull_request, which will trigger only once from a forked repo, and twice if from a separate branch of the same repo (which wasn't too bad, I guess?).

lu-wang-dl: What do you think about this workflow: Limit the push event to master only. So when any PR is merged, it will trigger one tests. But the pushes from  other branches will not trigger any test unless you make a PR.

yxiong: Sounds good. Updated to the following. PTAL.

```
on:
  push:
    branches:
        - master
  pull_request:
  workflow_dispatch:
```

wenfeiy-db (approved):
LGTM!

